PR #180, opened 2014-07-18
WYSIWYG comments / forums

jbarrett:
Finishing off the initial SCEditor integration work.

Creating new thread:
Creating new thread

Viewing new thread:
Viewing new thread

Replying:
Replying

Outstanding issues:

- Textarea width not filling containing div.
- Font should be sans-serif to match more closely what will be rendered
- Content load shield kills images

@sdougbrown might pick your brains on this later?

@zekiel you can check out the current state of affairs on https://view.dukgo.com/ (alongside the last set of password form changes)


jbarrett:
This is probably a redesign task - either way, this PR can't be merged sensibly after static file build stuff has been put in place.

PR #298, opened 2014-11-11
Experiments with postgres fulltext search

jbarrett:
We currently use dezi, but we already had a fulltext engine so why not use it?

Reasoning:
- Robustness
- KISS


jbarrett:
Closing this, rethinking.

PR #419, opened 2014-12-22
Prettify IAPage

jagtalon:
Here's what we need:
ia-single-v3


jagtalon:
@russellholt Ah thanks--I understand now. So just hide the GitHub files then?

russellholt:
No, not just the github files. it was done correctly before. Let's just remove the show more altogether.

jagtalon:
@russellholt alright gotcha thanks!

PR #428, opened 2015-01-03
IAPage: Remove "Show More" for now.

jagtalon:
CC @russellholt as discussed in https://github.com/duckduckgo/community-platform/pull/419#issuecomment-68543391


russellholt:
tested, LGTM. we can add some of this back later when/if we break files out into subsections.

PR #430, opened 2015-01-06
IAPage: Don't automatically add periods.

jagtalon:
Let's just depend on the metadata instead of adding a period automatically in the description.


russellholt:
:+1:

PR #431, opened 2015-01-06
IAIndex: Remove the topic class in the filter list.

jagtalon:
It's messing up the style.

screen shot 2015-01-06 at 10 37 48 am


russellholt:
ok

PR #432, opened 2015-01-06
IAIndex: Show the number of items in the header.

jagtalon:
Show the exact number of IAs in the list.

### DIFF
--- a/src/ia/js/IAIndex.js
+++ b/src/ia/js/IAIndex.js
@@ -30,7 +30,6 @@
             var query = ""; 
 
             $.getJSON(url, function(x) { 
-                $("#ia_index_header h2").text(x.length + " Instant Answers");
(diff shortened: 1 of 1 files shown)


src/ia/js/IAIndex.js
```diff
@@ -271,6 +270,8 @@
             var selector_all;
             var text_all;
             var tot_count = 0;
+
+            $("#ia_index_header h2").text("Showing " + $(".ia-item:visible").length + " Instant Answers");
```
russellholt: why not give an id for the h2 to address it directly? seems superfluous to search for it within the #ia_index_header.

russellholt: Actually we can look at things like that in a perf pass later. would rather move forward with functionality right now.

PR #434, opened 2015-01-06
IAIndex: Remove the "IsAwesome" IAs.

jagtalon:
There are a lot of IsAwesome IAs, and we might not want to showcase them in the index. Removing them from the list while we figure out what to do with them.

### DIFF
--- a/src/ia/js/IAIndex.js
+++ b/src/ia/js/IAIndex.js
@@ -4,6 +4,17 @@
         this.init();
     };
 
+    Handlebars.registerHelper('skipTest', function(ia, options) {
+        var topics = ia.instant_answer_topics;
+        for(var i = 0; i < topics.length; i++) {
+            if(topics[i].topic.name === "test") {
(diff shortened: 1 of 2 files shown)


russellholt:
use the test topic instead

jagtalon:
@russellholt updated!

russellholt:
do you really think grep is the best way to do this? why are we going through the list so many times in so many separate contexts?

jagtalon:
@russellholt made it simpler by adding a `skipTest` helper function that prevents the `li` tag from actually being added. I first started with just hiding the tag, but it ends up being more complicated because clearing the filter could show it again. So best to not put the `<li>` tag in there in the first place.

russellholt:
added WIP label because seems to be not needed, but I want to keep it around until I know we don't need it :8ball:

jagtalon:
@russellholt looks like we don't really need it. Closing.

PR #435, opened 2015-01-06
IAPage: Switch Github Files and Issues around.

jagtalon:
Switch the two around for aesthetic reasons.


src/templates/devinfo.handlebars
```diff
@@ -37,6 +18,24 @@
   </div>
 </div>
 
+<div class="ia-single--info">
+  {{#if code}}
+    <div class="ia-single--header">GITHUB FILES</div>
```
russellholt: Can we rename this to be `Source` ?

jagtalon: Sure thing

russellholt:
LGTM

Issue #426, opened 2014-12-31
Feedback icon font 404


jbarrett:
Not sure what changed, but it looks like the icon fonts for the feedback pages has gone AWOL:

![Feedback - missing icons](http://jbrt.org/ddg/feedback.png)

Checked so far:
- nginx, serving ddgc_static correctly
- feedback.css changes - none recently in this area.

Not sure when this was introduced.

cc @malbin @jdorweiler @sdougbrown @zekiel Thanks!

javathunderman:
@jbarrett Should this be closed?

zekiel:
believe so based on #440  and testing

PR #440, opened 2015-01-07
Changing font URLs to absolute path.

jbarrett:
#426


jbarrett:
@jdorweiler @russellholt I presume this change needs a `grunt build / release` after merge? Or are they done within the PR before merging?

PR #441, opened 2015-01-07
Adding grunt-available-tasks and grunt-bump to package.json

jbarrett:
@jdorweiler @russellholt Please let me know if the version pinning is appropriate.


jdorweiler:
@jbarrett  version looks good.  The version in package.json is always 1 higher than whatever it live.  
Thanks for adding the dependencies.  I must have forgot to add a `--save-dev` when I installed them.

jdorweiler:
I see you're asking about the package versions.  Those look good too :)

PR #442, opened 2015-01-07
IAIndex: Adjusted the arrow position

jagtalon:
It was too low in Firefox.


russellholt:
this is actually worse. just do text-align:center and it's fine. I think the :before stuff is way too complicated.

jagtalon:
@russellholt much better now!

screen shot 2015-01-08 at 1 42 33 pm

screen shot 2015-01-08 at 1 42 53 pm

russellholt:
way betta

:+1:

PR #444, opened 2015-01-08
IAPage: Choose Github over other services.

jagtalon:
This PR does two things:
- Chooses Github
- Completes the URL if all we have is the handle of the service.

### DIFF
--- a/src/ia/js/IAPage.js
+++ b/src/ia/js/IAPage.js
@@ -2,6 +2,31 @@
     // Handlebars helpers
     Handlebars.registerHelper('encodeURIComponent', encodeURIComponent);
 
+    // Choose a GitHub location before anything else.
+    Handlebars.registerHelper('chooseService', function(services) {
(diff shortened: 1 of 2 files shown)


src/ia/js/IAPage.js
```diff
@@ -2,6 +2,31 @@
     // Handlebars helpers
     Handlebars.registerHelper('encodeURIComponent', encodeURIComponent);
 
+    // Choose a GitHub location before anything else.
+    Handlebars.registerHelper('chooseService', function(services) {
+        // Make sure to build the URL if we just find the handle.
```
russellholt: check for null on services

jagtalon: @russellholt I don't think `services` can be null because `{{each}}` is already going through the array. So if `attribution` is null, `{{chooseService}}` won't be executed.

russellholt:
LGTM.

PR #445, opened 2015-01-08
IAPage: Hide the image's red bar, and create one in HTML.

jagtalon:
Here's how it looked:
screen shot 2015-01-08 at 4 20 01 pm

Here's how it looks now:
screen shot 2015-01-08 at 4 20 19 pm

What I did is hide the top 4px of the image, and replaced it with a `div` with the same color. Now it looks a little better.


russellholt:
good until/unless we trim the top of the image

PR #449, opened 2015-01-09
add more cascade options

jdorweiler:
@russellholt @MariagraziaAlastra @jbarrett  These two also need the cascade option turned on.  You will have to do the ddgc_db_autoupgrade after this is merged though.


MariagraziaAlastra:
@jdorweiler LGTM :+1:

PR #450, opened 2015-01-12
IA Pages Index - Fix the count for "All Topics"

MariagraziaAlastra:
@russellholt @jagtalon Now it displays the correct count. 
I've tested it, but I'd like you to test it too. Let's make sure there are no edge cases.


jagtalon:
@MariagraziaAlastra Works well! :+1:

PR #452, opened 2015-01-12
Db transactions

jdorweiler:
This fixes two things:
- **Better error logging**  
 - If there is something wrong with the metadata file it logs the actual error so we know what happened

- **Transactions**
 - All database updates/deletes are in a single transaction.  This means if anything fails the whole transaction will roll back

@russellholt @jbarrett @MariagraziaAlastra


MariagraziaAlastra:
@jdorweiler Nice!

script/iameta.pl
```diff
@@ -30,114 +28,113 @@
...
-    #     print color 'reset';
-    # }
-    # exit 1 if (++$line > 5);
-
-
-}
+};
 
+try{
+    $d->db->txn_do( $update );
```
russellholt: so does this commit for the entire operation?

jdorweiler: yeah.  It's all or nothing.  Any errors will cause a rollback.

russellholt:
LGTM

PR #453, opened 2015-01-12
IAIndex: Fix IE 8 errors.

jagtalon:
String.prototype.trim() is only available in IE 9, I think. Replacing them with `$.trim` fixes the problem.

screen shot 2015-01-12 at 4 23 45 pm

### DIFF
--- a/src/ia/js/IAIndex.js
+++ b/src/ia/js/IAIndex.js
@@ -73,7 +73,7 @@
 
                 if (((evt.type === "keypress" && evt.which === 13) && $(this).hasClass("text"))
                     || (evt.type === "click" && $(this).hasClass("filters--search-button"))) {
-                    var temp_query = $input_query.val().trim();
+                    var temp_query = $.trim($input_query.val());
                     if (temp_query !== query) {
                         query = temp_query;
                         ind.filter($list_item, query);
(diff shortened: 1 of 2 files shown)


russellholt:
great job fixing this.

so my only question is how much of a burden does the polyfill $.trim  put on the rest of us (the majority users) just to support IE8.

MariagraziaAlastra:
Besides, have we got any updated stats on IE 8 usage for DDG? I keep hoping that we can stop supporting it, soon enough...

jagtalon:
It's one line of regexp https://github.com/jquery/jquery/blob/0ea8c32863af31fb5cfc184e8d513bbae35583e8/src/core.js#L26 although the native one might be faster. Probably not much difference.

MariagraziaAlastra:
@jagtalon Looks like it's not completely fixed:
image

Steps to reproduce: load the Index page and click on any filter without changing the text inside the searchbar.
After clicking on "clear filters" the issue doesn't show up anymore:
image

IE 8 takes the placeholder text for the input field as an actual value, so when at first you click on a filter it applies that + the string "Filter Instant Answers", which of course doesn't produce any result.

And, in fact, it looks like [the placeholder attribute isn't supported neither in IE 8 nor in IE 9](https://stackoverflow.com/questions/5522164/input-placeholders-for-internet-explorer).

jagtalon:
@MariagraziaAlastra nice catch!

MariagraziaAlastra:
Thanks @jagtalon, this is ready to be merged now! :+1:

Issue #437, opened 2015-01-07
Prosody Integration and account handling issues


jbarrett:
- Some UTF-8 characters in the password field are troublesome (try creating an account with '€' in the password).
- Auth is done over HTTP. While both services are on the same host, this isn't a huge deal. We should switch to XMPP or HTTPS.
- #363 Deleted users not completely deleted

I suspect our UTF-8 issues would be resolved by using an XMPP module to register the Prosody account. Depending on how long this tends to take, we could move it to a job queue of sorts (we do a lot of emailing and indexing within HTTP sessions which could happen after - see #270).

It should be possible to do XMPP authentication without performing much of the additional [session lifetime](https://metacpan.org/pod/Protocol::XMPP#IMPLEMENTATION) elements - buddy list retrieval and such should be skipped.

While there is a fix for #363, other considerations around content deletion need to be decided (Comments, threads, translations, IA ideas, IA page permissions, attributions etc., account details...)

jbarrett:
There is a user account called 'deleted' which is configured in `DDGC->config->deleted_user`.

What currently occurs on user deletion:

https://github.com/duckduckgo/community-platform/blob/bd6ce62/lib/DDGC.pm#L711-L764

- Translation contributions user is set to 'deleted'
- Comments content is replaced and user is set to 'deleted'
- Threads are ghosted and user is set to 'deleted'

Taking a look at the User model's relationships should inform what else we should be handling:

- Data in `users` table itself - should lose everything but the username (to prevent duplicate registrations later)
- IA ideas
- Blog articles
- Screenshots? Avatars?
- IA page editing rights? Attribution? (Are there other relationships with IA pages?)

Some suggestions for IA attribution in #456 

ping @jdorweiler @zekiel @russellholt @MariagraziaAlastra

MariagraziaAlastra:
@jbarrett there's also the [assignee/sponsor for the IA Page](https://github.com/duckduckgo/community-platform/blob/dev/lib/DDGC/DB/Result/InstantAnswer.pm#L182) - this would be an admin watching over any activity and committing edits, adding metadata etc.

jbarrett:
What's actually stored in that column? If it's a DDGC user, it should refer to it with `users_id` (though one of the translation system tables uses username...)

MariagraziaAlastra:
I guess this uses username as well since it features 'text' as data-type. Maybe it would be more convenient to have the users_id, though. 
[...]

Issue #363, opened 2014-12-11
Deleted users not completely deleted


FantomskaFirma:
Forums view: https://view.dukgo.com/forum/general
Thread view: https://view.dukgo.com/forum/thread/3330/poc

👻 http://i.imgur.com/Eml8THA.png

jbarrett:
Thanks for reporting this @FantomskaFirma 

This should be fixed in an upcoming release, though it is currently staged on https://view.dukgo.com/ should you wish to check it out.

jbarrett:
While further discussion is required on specifics of account handling, this particular issue should be resolved.

Issue #325, opened 2014-11-24
Don't bump forum posts when comments set to deleted user


jbarrett:
During account deletion, comment contributions are set to a "deleted user" account (I don't like that there has to be a 'deleted user' account, but that's a separate issue). For each comment encountered, its 'updated' timestamp is reset which bumps the associated thread up the list on the forum.

*edit* A related issue is threads don't have their user reset - only the related comments.

PR #456, opened 2015-01-14
Fixing user deletion

jbarrett:
While the robustness of our Prosody "integration" needs to be evaluated (#437), we can provide some fixes for the brokenness right now. This should address:

#363 - Deleted users not completely deleted 

Prosody usernames are made lowercase prior to storing them, so we should select on a lowercase username.

#325 - Don't bump forum posts when comments set to deleted user

Explicitly setting the updated date to equal created prevents this.

### DIFF
--- a/lib/DDGC.pm
+++ b/lib/DDGC.pm
@@ -745,7 +745,7 @@ sub delete_user {
 		if ($self->config->prosody_running) {
 			$self->xmpp->_prosody->_db->resultset('Prosody')->search({
 				host => $self->config->prosody_userhost,
-				user => $username,
+				user => lc($username),
 			})->delete;
 		}
 	}
--- a/lib/DDGC.pm
+++ b/lib/DDGC.pm
@@ -744,6 +744,7 @@ sub delete_user {
 		my @threads = $user->threads->search({})->all;
 		for (@threads) {
 			$_->users_id($deleted_user->id);
+			$_->ghosted(1);
 			$_->update({ 'updated' => $_->created });
 		}
 		$guard->commit;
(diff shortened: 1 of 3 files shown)


jbarrett:
@zekiel This currently just sets the thread's user to 'deleted'.

Setting the thread title to "This user has been deleted" is not useful. I guess we could add it to the hidden pile.

jbarrett:
http://view.dukgo.com/

zekiel:
login is failing appropriately after deletion. 

@jbarrett  doesn't appear to work for IA ideas. GenRamblings will set to deleted user but Ideas remain w/ author and content.

jbarrett:
@russellholt @MariagraziaAlastra @jdorweiler What should happen to IA attribution etc. when a user deletes their account?

@zekiel Each contribution type needs explicit and specific handling which has to be decided, so every new thing that gets added, someone has to know to handle it in the delete code... :)

MariagraziaAlastra:
@jbarrett This is open for discussion, afaik we didn't think about this possibility.
Imho if the user created any IAs he should be asked whether he would like to keep its name without the link to the profile page, or use the github account for the link, or just have every reference to him removed.

In case he doesn't want his name at all on IA Pages either we could display something like "info not available" for the attribution...

russellholt:
I think @MariagraziaAlastra is right about the first part, the user should be given an option to change their attribution - which needs to be subject to review by a human.

But I think all users should have the option to be excluded from attribution, where there should just not be a listing for them. if it's blank I'm not sure what to say but I think 'no info available' sounds like an error.

jbarrett:
Would anyone object if we merged this fix for now and moved the discussion of account handling over to #437?

MariagraziaAlastra:
@jbarrett Sure, go ahead!

PR #457, opened 2015-01-14
IE 8 and IE 9 don't support the placeholder attribute

MariagraziaAlastra:
@jagtalon see my latest comment on #453.

Screen:
image


jagtalon:
:+1:

PR #462, opened 2015-01-19
IA Page - show latest edits if the user has the right permissions

MariagraziaAlastra:
... If not, show the live version instead.

Overview of the changes: the JSON endpoint for the single page now is composed of two hashes: `live` and `edited` (this last one only if the user has edit permissions or is an admin).
The `live` hash has, of course, the live version for each field, while the `edited` hash has the latest edit for each field, or, in case that field wasn't edited, the live version.

In IAPage.js the response from the JSON endpoint gets memorized in a variable, `ia_data`, and if this variable contains the `edited` hash then it means the user is either admin or has edit permissions, so we set `x` to `ia_data.edited`; otherwise we set it to `ia_data.live`. 

//cc @russellholt @jdorweiler @jagtalon @jbarrett

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -165,9 +165,11 @@ sub ia_json :Chained('ia_base') :PathPart('json') :Args(0) {
 
     my $ia = $c->stash->{ia};
     my @topics = map { $_->name} $ia->topics;
-
+    my $edited;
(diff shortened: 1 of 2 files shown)


MariagraziaAlastra:
Given that these are quite straightforward changes, I'm leaving this PR around until tomorrow and then merge, so speak now if you have questions, suggestions and/or critiques :)

russellholt:
LGTM.

PR #467, opened 2015-01-23
IA Pages - bring back page editing

MariagraziaAlastra:
After a massive refactoring, page editing is back.
The layout is totally different and independent from the static page one, but it uses the same principle as before: when clicking on the edit button, the page gets replaced by the pre_edit Handlebars templates, one for each editable field, shaping what I like to call the "Diff View", in which you can see the live version vs the edited one for each field:
screen shot 2015-01-23 at 14 37 25

To edit a field, just click on "Edit" at the top-right corner of the row, and the entire row will be highlighted in green, and under the "Edited" column will show up an input containing either the latest edit for that field, if available, otherwise the live version. To save just hit "return" on the keyboard or click on "Done" on the top-right corner of the row:
screen shot 2015-01-23 at 14 38 00
screen shot 2015-01-23 at 14 38 05

Topics and examples have a different layout. I kept the add and remove buttons, arranged differently, and it saves only if you click on "Done":
screen shot 2015-01-23 at 14 42 25

The values are saved if and only if they are different both from the latest edit and from the live version, otherwise the edit_template gets replaced by the pre_edit one, but without changes. Notice that the edit templates only contain the elements for the edited column for each field instead of the whole row.

I've also done some code refactoring, removed unused code, organized the existing one better, and I added a function inside InstantAnswer.pm which removes the whole updates column after the commit, without the need of looping through the edits keys.

That's it, I think I said pretty much everything. Let me know what you think!

//cc @russellholt @jagtalon @jdorweiler @jbarrett

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -123,29 +123,29 @@ sub ia_base :Chained('base') :PathPart('view') :CaptureArgs(1) {  # /ia/view/cal
 
     my $permissions;
     my $is_admin;
-    my $edit_class = "hide";
-    my $commit_class = "hide";
+    my $can_edit;
+    my $can_commit;
 
     if ($c->user) {
         $permissions = $c->stash->{ia}->users->find($c->user->id);
         $is_admin = $c->user->admin;
-    }
 
-    if ($permissions || $is_admin) {
-        $edit_class = "";
+        if ($permissions || $is_admin) {
+            $can_edit = 1;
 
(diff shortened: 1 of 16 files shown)


MariagraziaAlastra:
- [Demo screencast](http://gfycat.com/FatAllGreatdane#?speed=2)
- [Screencast 2 - edit topics](http://gfycat.com/DevotedCluelessEgg#?speed=2)

russellholt:
ok looking good

for the record the plan is to merge this and then do a couple cleanup passes for layout issues.

PR #470, opened 2015-01-26
IA Page: Prevent text from overlapping.

jagtalon:
CC @MariagraziaAlastra 

Old:
screen shot 2015-01-26 at 12 17 02 pm

New:
screen shot 2015-01-26 at 12 17 23 pm

It could use some help distinguishing the individual queries, but it works and we can circle back to that.


MariagraziaAlastra:
:+1:

PR #471, opened 2015-01-26
IA Pages - prevent adding duplicated topics (and queries)

MariagraziaAlastra:
@jagtalon Turns out it was an issue with trimming the text - some values can have spaces or newlines at the end and aren't recognized as the same topic!

While I was working on this, I fixed other queries as well :)


jagtalon:
Nice!

PR #475, opened 2015-01-27
IA Pages - add toggle-view button to switch between live and edited data

MariagraziaAlastra:
@russellholt @jagtalon 

Default view for logged in users with edit permissions:
screen shot 2015-01-27 at 17 52 08

After clicking on "live" (notice how name and description values change):
screen shot 2015-01-27 at 17 52 12

### DIFF
--- a/src/ia/js/IAPage.js
+++ b/src/ia/js/IAPage.js
@@ -82,6 +82,7 @@
 
                         $("#edit_disable").removeClass("hide");
                         $(this).hide();
+                        $(".special-permissions__toggle-view").hide();
                     });
 
                     $(".special-permissions__toggle-view__button").on('click', function(evt) {
--- a/src/ia/js/IAPage.js
+++ b/src/ia/js/IAPage.js
@@ -47,9 +47,7 @@
                     // Show latest edits for admins and users with edit permissions
                     var x = {};
                     if (ia_data.edited) {
-                        $.each(ia_data.live, function(key, value) {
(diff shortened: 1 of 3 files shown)


jagtalon:
@MariagraziaAlastra Should "Live | Edited" disappear when there's nothing to see? Kinda like how "Commit edits" shows up and disappears

src/ia/js/IAPage.js
```diff
@@ -242,14 +261,26 @@
             'other_queries'
         ],
 
+        updateData: function(ia_data, x, edited) {
+            $.each(ia_data.live, function(key, value) {
+                if (edited) {
+                    x[key] = ia_data.edited[key]? ia_data.edited[key] : value;
+                } else {
+                    x[key] = value;
```
jagtalon: Is it possible to just return `ia_data.live` if `edited` is `false` so that we don't have to go through the whole thing again? Not sure if I'm understanding it correctly, but it looks like `ia_data.live` is the same as `x` when `edited` is `false`.

MariagraziaAlastra:
@jagtalon Good idea!

src/ia/js/IAPage.js
MariagraziaAlastra: @jagtalon Thing is, directly assigning ia_data.live to x clones the object, it doesn't just copy it. So if the user later switches again to edited and x changes again, ia_data.live will be changed as well.

jagtalon: Ah thanks for that.

jagtalon:
:+1:

Issue #476, opened 2015-01-27
Schema migration policy.


jbarrett:
[`script/ddgc_db_autoupgrade.pl`](https://github.com/duckduckgo/community-platform/blob/dev/script/ddgc_db_autoupgrade.pl) should be considered informational for developers working on schema changes, not the canonical source of diffs post-merge/deployment.

I'd like to discuss options for making schema changes available to developers and for deployment. High-level proposal, we have a `sql/` or `migration/` directory which contains a script containing DDL and data migration for each task or version. These scripts can then be run on every required instance, plus we have a reasonably easy-to-reference history of changes.

So, say the last version was 1.006, in preparing for version 1.007 we would have:

`sql/1.007/task-branchname.sql`
`sql/1.007/other-task-branchname.sql`

...containing the required DDL and data updates for that version - you bear the responsibility for defining the DDL when schema changes are made. I'm not sure if the idea needs much expansion beyond this :)

@jdorweiler @MariagraziaAlastra @russellholt Thanks!

MariagraziaAlastra:
@jbarrett So what should we do exactly when we make schema changes? If I'm understanding correctly, we should create a new sql file for the new version containing all the needed updates, and then execute it, right?

jbarrett:
Yep, pretty much - there should be an easy way for developers who merged your changes to sync their postgres instance and it could also act as an audit trail.

Yes, you could look at the history of `DDGC::DB::*`, but the exact DDL and index changes would be hard to derive from this.

MariagraziaAlastra:
@jbarrett Cool, seems reasonable to me.

jbarrett:
I guess we'll leave it up to the next person making database changes to set the policy :)

jdorweiler:
I'm the lucky first victim :cactus:   
It sounds like I should run `ddgc_schema_diff` and commit the output from that script into `sql/version/`?  Then someone would still have to manually update their DB using that file.

jbarrett:
`ddgc_schema_diff` produces a lot of erroneous extra output - ideally, if you were to use it, you would at least prune this.

But yeah, that's the gist :)

PR #480, opened 2015-01-28
Renaming ddgc_db_autoupgrade

jbarrett:
This isn't a feasible way to upgrade db schemas - the diff should be considered informational, not functional:

- Renamed the script and removed the `--doupgrade` option.

Output from this script can be used to inform DDL scripts for #476.


MariagraziaAlastra:
@jbarrett Tested, LGTM :+1:

PR #483, opened 2015-01-29
IAPage: Updated create new issue link target

mrshu:
- Updated the target of create new issue link so that it looks a bit
  better and at the same time encourages writing a title.

Signed-off-by: mr.Shu mr@shu.io

---

The previous link already had a default title and so it seemed that one would not be too interested in writing one.

Moreover, I think that the link back to the IAPage should be a bit more distinguished in the issue message.

### DIFF
--- a/src/templates/devinfo.handlebars
+++ b/src/templates/devinfo.handlebars
@@ -11,7 +11,7 @@
   {{/if}}
 
   <div class="create-issue">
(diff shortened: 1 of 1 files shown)


mrshu:
to @jagtalon 

cc @MariagraziaAlastra

russellholt:
:+1:

MariagraziaAlastra:
@mrshu Thanks! :+1:

PR #484, opened 2015-01-29
ghissues regex bug and added transactions 

jdorweiler:
The delete/update is done in a transaction so it will roll back on any errors.  
@MariagraziaAlastra @russellholt @jbarrett

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -7,7 +7,7 @@ use lib $FindBin::Dir . "/../lib";
 use JSON;
 use DDGC;
 use Data::Dumper;
-
+use Try::Tiny;
 my $d = DDGC->new;
 
 # JSON response from GH API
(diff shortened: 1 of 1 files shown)


MariagraziaAlastra:
Uhm, I tried running this on my instance and the new issues in which I added the IA Page link aren't showing up... plus I get this message:

```
HTTP::Server::PSGI: Accepting connections at http://0:5000/
DBIx::Class::Storage::DBI::select_single(): Query returned more than one row.  S                                                                                        QL that returns multiple rows is DEPRECATED for ->find and ->single at /home/ubu                                                                                        ntu/community-platform/script/../lib/DDGC/Web/Controller/InstantAnswer.pm line 1                                                                                        72
```

MariagraziaAlastra:
Uhm, so it looks like the problem might be [line 172 of InstantAnswer.pm](https://github.com/duckduckgo/community-platform/blob/dev/lib/DDGC/Web/Controller/InstantAnswer.pm#L172), because "find" gets just one row, but we want to retrieve multiple ones.

MariagraziaAlastra:
Fixed that error in #485, but I still don't see the latest issues I added. For example [this one for AlternativeTo](https://github.com/duckduckgo/zeroclickinfo-spice/issues/1305) isn't showing up.

jdorweiler:
@MariagraziaAlastra Thanks for checking it out.  Looks like the api only gives 30 issues at a time so I'll have to update it to get multiple pages.

jdorweiler:
okay,  now with fewer bugs!

MariagraziaAlastra:
@jdorweiler haha, thank you! Works like a charm now :+1:

PR #485, opened 2015-01-29
InstantAnswer.pm - use search() to get multiple rows for issues

MariagraziaAlastra:
See comments on #484 
@russellholt @jdorweiler @jbarrett


russellholt:
:+1:

PR #489, opened 2015-02-03
IA Pages - add dev milestone row in edit mode if user is an admin

MariagraziaAlastra:
The dev milestone value can be edited using a dropdown (the available options are hardcoded in the Handlebars template for now, until we are sure about what they should be), and, since the commit page has been updated as well, the edits to this field can be committed.

//cc @russellholt @jagtalon 
screen shot 2015-02-03 at 15 31 15
screen shot 2015-02-03 at 15 37 33


jagtalon:
:+1: @MariagraziaAlastra

jagtalon:
@MariagraziaAlastra completely missed this but there are two "Planning"s

screen shot 2015-02-03 at 2 03 22 pm

MariagraziaAlastra:
@jagtalon It's the same for topics! I think it should be fairly easy to hide the duplicated list item though, I'll make a task for that :)

PR #490, opened 2015-02-03
IAPages: Align the lists

jagtalon:
Here's how it used to look:

screen shot 2015-02-03 at 1 32 24 pm

If you notice, the live columns in "Other Queries" and "Topics" are not aligned properly. One solution is to vertically align them so that the list is centered in the middle:

img_0010

But it was a bit difficult to pull off. So to put in the least amount of changes, I did this instead:

screen shot 2015-02-03 at 1 30 51 pm
screen shot 2015-02-03 at 1 39 23 pm

I think it does the job, but tell me what you think @MariagraziaAlastra!


jagtalon:
screen shot 2015-02-03 at 2 06 19 pm

MariagraziaAlastra:
@jagtalon I think it looks great!

MariagraziaAlastra:
Tested, LGTM :+1:

PR #492, opened 2015-02-04
IA Pages - load different templates on the static page based on the dev milestone

MariagraziaAlastra:
WIP, will add Handlebars templates afterwards; @jagtalon will help, so he's going to make some PRs to this branch.
//cc @russellholt

### DIFF
--- a/src/ia/js/IAPage.js
+++ b/src/ia/js/IAPage.js
@@ -373,6 +373,7 @@
                     $(".ia-single--left").append(templates.live[this.field_order[i]]);
                 }
 
+                $(".ia-single--right").before(templates.live.name);
                 $(".ia-single--right").append(templates.live.screens);
 
                 $(".show-more").click(function(e) {
@@ -389,7 +390,7 @@
                     }
                 });
             } else {
-                $(".ia-single--left, .ia-single--right").hide();
+                $(".ia-single--left, .ia-single--right, .ia-single--name").hide();
                 if (dev_milestone === "live") {
                     $(".ia-single--edits").removeClass("hide");
                     for (var i = 0; i < this.edit_field_order.length; i++) {
--- a/src/ia/js/IAPage.js
+++ b/src/ia/js/IAPage.js
@@ -45,24 +45,30 @@
                 $.getJSON("/ia/view/" + DDH_iaid + "/json", function(ia_data) {
 
                     // Show latest edits for admins and users with edit permissions
-                    var x = {};
+                    var latest_edits_data = {};
                     if (ia_data.edited) {
-                        x = DDH.IAPage.prototype.updateData(ia_data, x, true);
+                        latest_edits_data = DDH.IAPage.prototype.updateData(ia_data, latest_edits_data, true);
                    } else {
-                        x = ia_data.live;
+                        latest_edits_data = ia_data.live;
                     }
 
                     // Readonly mode templates
                     var readonly_templates = {
-                        name : Handlebars.templates.name(x),
-                        status : Handlebars.templates.status(x),
-                        description : Handlebars.templates.description(x),
-                        topic : Handlebars.templates.topic(x),
-                        screens : Handlebars.templates.screens(x),
-                        template : Handlebars.templates.template(x),
-                        examples : Handlebars.templates.examples(x),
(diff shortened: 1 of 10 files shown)


jagtalon:
@MariagraziaAlastra Added more space for the title and added the live label:

screen shot 2015-02-05 at 2 13 50 pm

MariagraziaAlastra:
@jagtalon Awesome, thanks!

jagtalon:
It should probably align with the "Enable edit mode" though

MariagraziaAlastra:
@jagtalon Actually not, for now let's leave it like this, there's a lot of stuff going on on the top right corner and I fear aligning the IA name with that would clump a bit the view. Anyway we can always get back to that :)

jagtalon:
@MariagraziaAlastra oh yeah def that's why I didn't do anything :)

jagtalon:
@MariagraziaAlastra about the "Safe Search"--it's more of a "code" thing than a "metadata" thing isn't it? That is, toggling this would mean changing the behavior of https://duckduckgo.com/?q=ud+brb&kp=-1&ia=dictionary and not just the data about it.

I think I would expect it to change an IA from "Safe" to "Unsafe" and vise versa which we wouldn't want to do, right?

MariagraziaAlastra:
@jagtalon No, that's not what's going to do, it's just informational. Like, does it require the safe search to be off or not? And then the developer will code accordingly :)

jagtalon:
@MariagraziaAlastra oh right because it's the planning stage! Thanks

src/ia/js/IAPage.js
```diff
@@ -45,24 +45,53 @@
                 $.getJSON("/ia/view/" + DDH_iaid + "/json", function(ia_data) {
 
                     // Show latest edits for admins and users with edit permissions
-                    var x = {};
+                    var latest_edits_data = {};
                     if (ia_data.edited) {
-                        x = DDH.IAPage.prototype.updateData(ia_data, x, true);
-                   } else {
-                        x = ia_data.live;
+                        latest_edits_data = DDH.IAPage.prototype.updateData(ia_data, latest_edits_data, true);
```
russellholt: calling a function on the prototype seems kinda weird. No `new` ?

MariagraziaAlastra:
Still WIP, but on a good track. The functionality is there, the layout is there (thanks @jagtalon), we just need to decide exactly what fields to add and where, and then add them, of course :)

Dev page for users without permissions:
screen shot 2015-02-06 at 15 04 06


Dev page for admins (same thing for users with edit permissions, only without the "Mark as complete" button):
screen shot 2015-02-06 at 15 04 21

jagtalon:
@MariagraziaAlastra not sure if this is for this PR, but I think we should go with the repo being a dropdown instead.

For the Perl Module--would that be useful? What if there are multiple ones like Dictionary?

MariagraziaAlastra:
@jagtalon I guess that only the triggering one should be included there, since that's how we are distinguishing proper IAs from auxiliary submodules.

Yeah a dropdown would be better for that, not sure if I'm going to do that in this PR though

jagtalon:
@MariagraziaAlastra okay thanks we can do it in another PR. :) Hmm, yeah it's tricky like when someone makes a PR with multiple triggering submodules like SeatGeek. I guess we could just treat that as different IAs

MariagraziaAlastra:
@jagtalon Yup :)

jagtalon:
@MariagraziaAlastra Maybe "Triggers" would be useful in the dev phase.

MariagraziaAlastra:
Now the repo has a link to GitHub and there's also a link to the docs next to it :)
I wonder if we can link to the docs for other fields as well? Uhm.
screen shot 2015-02-06 at 16 40 04

@jagtalon Yeah triggers should definitely be in planning.

MariagraziaAlastra:
screen shot 2015-02-09 at 23 02 54

MariagraziaAlastra:
Final version (for now) - ready for review. Will add the new fields to the DB and make improvements in other PRs.
screen shot 2015-02-10 at 15 48 30

jagtalon:
:+1: Nice!

PR #500, opened 2015-02-11
IA Pages - add '&ia=[tab name]' to the example links

MariagraziaAlastra:
Not a big change, really. 
Just making sure that when a user clicks on an example link on the static page he gets the intended IA tab for that query.


MariagraziaAlastra:
Merging, since it's a minor change

PR #504, opened 2015-02-12
IA Pages - show the description in the dev page

MariagraziaAlastra:
@russellholt @jagtalon The description will be added alongside with name and id when a new IA page will be created, so we want to show the description not only when the IA is deployed and live, but also during the development process, in the dev page.

In order to accomplish this I had to separate topics and description, which were previously in the same Handlebars template.
I also fixed a little bug in InstantAnswer.pm - not all the IAs have a defined tab name, so it's necessary to check before trying to transform it to lowercase.


MariagraziaAlastra:
screen shot 2015-02-12 at 14 43 26

jagtalon:
:+1: Nice!

PR #506, opened 2015-02-12
IA Pages - front-end improvements

MariagraziaAlastra:
@russellholt @jagtalon 
While testing IA pages I noticed a couple of things which could be improved:
* On the dev page, the input fields are only saved on return keypress, but if you have to edit multiple fields it feels more natural to just click on the next one after each edit, so now it saves the values also on focusout (if the input had focus before).
* Comma-separated values were transformed into JSON strings even when the value was empty, so I added a check for that.
* The "clear filters" button on the Index doesn't show the correct cursor on hover, so I added `cursor: pointer;`


jagtalon:
@MariagraziaAlastra 

I went into the Perl Module and added "hi". Found this when I returned to the page:

screen shot 2015-02-12 at 11 33 02 am

Clicked on it but I didn't edit anything. Went back to the page:

screen shot 2015-02-12 at 11 34 13 am
screen shot 2015-02-12 at 11 34 28 am

jagtalon:
With the "Tab Name" I added "Software Stuff" but when I came back to it it just said "softwarestuff" (space went missing):

screen shot 2015-02-12 at 11 36 29 am

jagtalon:
Other than that it looks good!

MariagraziaAlastra:
@jagtalon ah yes, the Perl Dependencies thing is because it doesn't decode the JSON string before returning it on the endpoint. 
The tab name is because I'm getting it ready on the Perl side for adding it to the examples link, on the static page. Good catch, thanks!

jagtalon:
@MariagraziaAlastra oh okay then so you think that should go into another PR?

MariagraziaAlastra:
@jagtalon Nope, fixing now :)

jagtalon:
@MariagraziaAlastra nice it works! :+1:

PR #508, opened 2015-02-12
IA Pages - don't modify the IA tab string if dev milestone !== live

MariagraziaAlastra:
As said in the comments for #506. Now the tab name is modified (lowercase + spaces removed) inside IAPage.js, and only if the dev milestone is "live", since in that case we need to use it for the example links (the "&ia=[tab name]" bit).


MariagraziaAlastra:
Minor change, merging

PR #510, opened 2015-02-16
Hourly limit

jbarrett:
@zekiel, we'll need to decide on a text / icon to replace the placeholder here, I think:

placeholder


zekiel:
close out per IRL conversation. (at least, for now)

PR #514, opened 2015-02-16
Dev IA Page - add 'assign to me' buttons for team roles

MariagraziaAlastra:
@russellholt @jagtalon 

I'm logged in as "cpan", which is the specified username for the producer field, so the button is hidden for that one:
screen shot 2015-02-16 at 19 13 54


russellholt:
:+1:

jagtalon:
@MariagraziaAlastra woot!

PR #519, opened 2015-02-17
Jd/iameta fixes

jdorweiler:
Added a cascade delete for topics table when you run `iameta.pl delete`.  Switched from `decode_json` to `from_json` since it was causing problems with reading the metadata file. 

@russellholt @MariagraziaAlastra @jbarrett


MariagraziaAlastra:
@jdorweiler Thanks, LGTM! :+1:

PR #520, opened 2015-02-18
IA Pages - add 'New IA Page' button and form to the Index

MariagraziaAlastra:
@russellholt @jagtalon 
screen shot 2015-02-18 at 14 55 23
screen shot 2015-02-18 at 14 55 28
screen shot 2015-02-18 at 14 56 02
screen shot 2015-02-18 at 14 56 28

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -487,6 +487,37 @@ sub save_edit :Chained('base') :PathPart('save') :Args(0) {
     return $c->forward($c->view('JSON'));
 }
 
+sub create_ia :Chained('base') :PathPart('create') :Args() {
+    my ( $self, $c ) = @_;
+
+    my $ia = $d->rs('InstantAnswer')->find({lc id => $c->req->params->{id}});
+    my $is_admin;
+    my $result = '';
+
+    if ($c->user && (!$ia)) {
+       $is_admin = $c->user->admin;
+
+        if ($is_admin) {
+            my $new_ia = $d->rs('InstantAnswer')->create({
+                lc id => $c->req->params->{id},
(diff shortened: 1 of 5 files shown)


jagtalon:
@MariagraziaAlastra well that looks sleek

jagtalon:
@MariagraziaAlastra 
- Probably set the width of the textarea to 100%
screen shot 2015-02-18 at 12 52 50 pm
- Looks weird that the button disappears after you click on it.
- Shouldn't every new IA start off with planning and not anything else? i.e., why can I choose dev milestones?
- When I first click on it it looks like there's already some stuff in it.
screen shot 2015-02-18 at 12 57 02 pm
- I find that it's hard to get to because it's not in the list of IAs :( So if I put in the IA "Jag", then I search the filer it's not there. I know that we don't wanna put it there when it's not live but it's inconvenient to go to the URL bar and type in the ID that I want. This is, of course, outside the scope of this PR. :)

MariagraziaAlastra:
@jag Because if we want to port the PRs we have to the IA Pages then we won't probably mark them as "planning". Some of them are in development, some are in QA, some are just waiting to be deployed, I guess.

The button disappears because you don't need it after clicking on it, since you probably want to create just one IA at a time.

I could make it redirect to the new IA Page, anyway having the button in the IA Index is just a temporary solution to allow admins to add new IAs as soon as possible, that's why it's not really refined and everything

MariagraziaAlastra:
screen shot 2015-02-18 at 20 50 20

russellholt:
I think it's good. we clearly need the dev pipeline and it will be more clear how it should work once that's there. this is an essential start for now though.

MariagraziaAlastra:
Merging, thanks @jagtalon and @russellholt

jagtalon:
:+1:

PR #521, opened 2015-02-18
IA Page - fix bug with live-edited switch toggle

MariagraziaAlastra:
@jagtalon Tested and WFM, could you please confirm everything works as expected before merging?


jagtalon:
@MariagraziaAlastra WFM as well! :+1:

PR #523, opened 2015-02-18
examples.handlebars - fix bug in retrieving the tab for other queries

MariagraziaAlastra:
@jagtalon the "&ia=[tab name]" was already there for other queries too, just had to fix a bug with the Handlebars template: inside the loop it needs special syntax to access parent values.


jagtalon:
@MariagraziaAlastra :+1:

PR #527, opened 2015-02-19
Allow hiding of redundant help pages

jbarrett:
Set category to '-' to hide a help page. Content remains intact should we want to re-publish it later.

@zekiel thanks!


zekiel:
thank you much

PR #528, opened 2015-02-19
IA Pages Index - sort topics alphabetically in the topics filtering section

MariagraziaAlastra:
@russellholt @jagtalon I don't have the latest metadata yet, but I believe you can see that the list is sorted anyway :)
screen shot 2015-02-19 at 17 35 30


russellholt:
nice
is this better than doing it in javascript?

MariagraziaAlastra:
Yes I think so, thanks :)

jagtalon:
@MariagraziaAlastra noice!

PR #529, opened 2015-02-19
IA Pages creation - replace underscore with space for the status field

MariagraziaAlastra:
Before (look at the status): 
screen shot 2015-02-19 at 20 41 34

After:
screen shot 2015-02-19 at 20 40 15


jagtalon:
:+1:

PR #531, opened 2015-02-19
IA static Pages - fix bug with live-edited switch for status

MariagraziaAlastra:
@jagtalon 
screen shot 2015-02-19 at 22 03 15
screen shot 2015-02-19 at 22 03 17


jagtalon:
Nice! :+1:

PR #532, opened 2015-02-19
from and to json

jdorweiler:
`decode_json` fails when it runs into weird utf8 characters.  I think the long term fix is to use a separate updates table `One IA has Many updates` similar to the issues table.  That way we can avoid json all together.  I'm happy to work on that but here is a short term fix if we want it. 

@MariagraziaAlastra @jbarrett @russellholt


MariagraziaAlastra:
LGTM :+1:

PR #533, opened 2015-02-19
IAPage.js - change input value after clicking on 'assign to me'

MariagraziaAlastra:
@russellholt As discussed, no need to reload the page now.


russellholt:
LGTM

PR #537, opened 2015-02-23
Require action token for transactions with github.com

jbarrett:
Should, in effect, act as a state token as described here:

http://tools.ietf.org/html/draft-ietf-oauth-v2-20#section-10.12

@zekiel @malbin @nilnilnil Thanks!


nilnilnil:
:+1: nice, thanks. can you follow up and make sure the reporter gets a shirt?

PR #539, opened 2015-02-24
IA Pages - specific type-only fields (plus new Handlebars helpers file)

MariagraziaAlastra:
@russellholt @jagtalon 
We needed a generic Handlebars helpers file: from now on every helper will be defined there.
In this specific case I defined a block helper checking if two values are equal - it makes much easier to check for specific repos and can be used to refactor some other parts as well. I believe in the end having more trivial helpers like this one will save us from unnecessary overheads and ugly hacks, and will boost code performance as well. Now, screenshots!

Fathead shows src options field:
schermata 2015-02-24 alle 14 42 20

Spice shows API docs:
schermata 2015-02-24 alle 14 43 40

Other types don't show neither API docs, nor src options:
schermata 2015-02-24 alle 14 44 02

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -33,6 +33,7 @@ module.exports = function(grunt) {
     var ia_page_js = [
         'handlebars_tmp',
         'DDH.js',
+        'Helpers.js',
         'IADevPipeline.js',
         'IAIndex.js',
         'IAPage.js',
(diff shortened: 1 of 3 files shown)


MariagraziaAlastra:
Removed Fathead-specific section for now, since it needs a bunch of checkboxes: will do that in a separate PR.

PR #540, opened 2015-02-24
Dev Pages - multiple selection for topics

MariagraziaAlastra:
@russellholt @jagtalon 
I think three topics should be more than enough for an IA:
schermata 2015-02-24 alle 21 31 39

Removing a topic doesn't actually remove the select element, it just empties the first option value and sets that one as selected, and then saves the new list of topics:
schermata 2015-02-24 alle 21 31 51

### DIFF
--- a/src/ia/css/ia.css
+++ b/src/ia/css/ia.css
@@ -81,6 +81,7 @@
     padding-left: 0.2em;
 }
 .ia-single .row-diff .available_topics,
+.ia-single .topic-group,
 .ia-single .row-diff .editpage span input {
     border-top-left-radius: 0;
     -moz-border-top-left-radius: 0;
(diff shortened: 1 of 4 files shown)


MariagraziaAlastra:
If we decide we want to limit the number of other queries, too, I'll probably use the same approach on the editable page, since it seems more straightforward and less convoluted than removing the actual element and then having to click on the "+" button to get a new one.

russellholt:
@MariagraziaAlastra I agree about limiting the topics. makes it easier.

russellholt:
about other queries - what about just a text area, one query per line?

MariagraziaAlastra:
@russellholt That would mean more human error. You would have to explain "please each query on a newline" to everyone... I can already imagine the results

PR #542, opened 2015-02-25
Dev Pages - show Fathead-specific section

MariagraziaAlastra:
@russellholt data returned from the JSON endpoint after saving:
```json
"src_options":{
    "skip_end":0,
    "skip_abstract":1,
    "skip_qr":0,
    "skip_image_name":0,
    "skip_icon":0,
    "directory":"test",
    "min_abstract_length":"10",
    "skip_abstract_paren":1,
    "is_wikipedia":0,
    "src_info":"",
    "is_mediawiki":0
}
```

Screenshot:
schermata 2015-02-25 alle 19 26 46

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -324,7 +324,8 @@ sub ia_json :Chained('ia_base') :PathPart('json') :Args(0) {
                 mobile_ios => $ia->mobile_ios,
                 tested_relevancy => $ia->tested_relevancy,
                 tested_staging => $ia->tested_staging,
-                is_live => $ia->is_live
+                is_live => $ia->is_live,
+                src_options => $ia->src_options? from_json($ia->src_options) : undef
     };
 
     if ($c->user) {
--- a/src/ia/css/ia.css
+++ b/src/ia/css/ia.css
@@ -1195,6 +1195,10 @@ body.texture {
     margin-right: 0.3em;
 }
 
+.dev_milestone-container label {
+    width: 100%;
+}
+
 /* Set the height of the read-only fields so that 
    the layout doesn't change when it switches to the input field */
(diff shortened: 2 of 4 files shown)


russellholt:
works great, the only thing I'd change now is for `min_abstract_length` to be stored as a number

MariagraziaAlastra:
@russellholt Done :)

mrshu:
@russellholt sorry to interrupt, but `skip_qr` is supposed to be a regular expression and thus I'd expect it to be a string.

Other than that it looks great!

MariagraziaAlastra:
Ah darn, I thought it was a boolean. Thanks @mrshu!

mrshu:
@MariagraziaAlastra no problem at all, I do not see how you would know that =)

russellholt:
thanks @mrshu, did not know that.  grepping gives these examples:

"RuneScape|RuneScapeWiki|RS|Exchange"
"Star Wars Fanon|SWF|SWFanon"

so they're pretty simple

russellholt:
@mrshu do we still use skip_icon? they're all empty

mrshu:
@russellholt yes, they are, one would wonder whether they are regular expressions after all.

While they are rarely used in the DuckDuckHack fatehads (well, they are not used at all) they are strings and they should be represented as such.

Regarding `skip_icon`, that is an old artefact, sources that used them are not enabled at the moment. That being said, I think we could start using them again and so I think we should be ready to support it.

MariagraziaAlastra:
schermata 2015-02-25 alle 23 10 46
schermata 2015-02-25 alle 23 10 48

russellholt:
LGTM

mrshu:
Thanks @MariagraziaAlastra!

MariagraziaAlastra:
Thank you @mrshu @russellholt!

russellholt:
:cactus:

PR #547, opened 2015-02-26
Dev Page - bug fixes and improvements

MariagraziaAlastra:
@russellholt @jagtalon 
The main thing is that now you can delete values, and then there's some other fix like the test machine url one.


MariagraziaAlastra:
Link to duck.co user for team fields:
schermata 2015-02-26 alle 20 35 05

russellholt:
:+1:

PR #548, opened 2015-02-26
IA Pages - Handlebars helper for appending the tab name to the example queries

MariagraziaAlastra:
@jagtalon @russellholt So now we can have the tab name appended on the dev page example query, too!

### DIFF
--- a/src/templates/qa.handlebars
+++ b/src/templates/qa.handlebars
@@ -26,7 +26,7 @@
             {{#if permissions.can_edit}}
                 <input type="text" class="dev_milestone-container__body__input js-autocommit" id="example_query-input" value="{{live.example_query}}" />
             {{else}}
-                <a href="https://{{live.test_machine}}.duckduckgo.com/?q={{encodeURIComponent live.example_query}}{{tab_url live.tab}}" class="dev_milestone-container__body__readonly" id="example_query-a">
+                <a href="{{#if live.test_machine}}https://{{live.test_machine}}.duckduckgo.com/?q={{encodeURIComponent live.example_query}}{{tab_url live.tab}}{{/if}}" class="dev_milestone-container__body__readonly" id="example_query-a">
                     {{live.example_query}}
                 </a>
             {{/if}}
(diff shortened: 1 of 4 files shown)


src/templates/qa.handlebars
```diff
@@ -26,7 +26,7 @@
             {{#if permissions.can_edit}}
                 <input type="text" class="dev_milestone-container__body__input js-autocommit" id="example_query-input" value="{{live.example_query}}" />
             {{else}}
-                <a href="https://{{live.test_machine}}.duckduckgo.com/?q={{encodeURIComponent live.example_query}}" class="dev_milestone-container__body__readonly" id="example_query-a">
+                <a href="https://{{live.test_machine}}.duckduckgo.com/?q={{encodeURIComponent live.example_query}}{{tab_url live.tab}}" class="dev_milestone-container__body__readonly" id="example_query-a">
```
jagtalon: Should check if `live.test_machine` exists or else `http://.duckduckgo.com` is what's left.

src/templates/qa.handlebars
```diff
@@ -26,7 +26,7 @@
             {{#if permissions.can_edit}}
                 <input type="text" class="dev_milestone-container__body__input js-autocommit" id="example_query-input" value="{{live.example_query}}" />
             {{else}}
-                <a href="https://{{live.test_machine}}.duckduckgo.com/?q={{encodeURIComponent live.example_query}}" class="dev_milestone-container__body__readonly" id="example_query-a">
+                <a href="{{#if live.test_machine}}https://{{live.test_machine}}.duckduckgo.com/?q={{encodeURIComponent live.example_query}}{{tab_url live.tab}}{{/if}}" class="dev_milestone-container__body__readonly" id="example_query-a">
```
jagtalon: It looks kinda broken when I click on it and it doesn't go anywhere. Still better than the old one though :+1:

jagtalon: Would prefer something like 

```handlebars
{{#if live.test_machine}}
    <div class="dev_milestone-container__body__readonly" id="test_machine-txt">
        {{live.example_query}}
    </div>
{{else}}
     <a href="...
{{/if}}
```

jagtalon:
:+1:

PR #551, opened 2015-02-27
Helpers.js - check tab is defined before checking length

MariagraziaAlastra:
@russellholt @jdorweiler 
schermata 2015-02-27 alle 17 28 49


russellholt:
:+1:

PR #552, opened 2015-03-02
Increased contrast in header text to pass WCAG 2

tagawa:
Slightly increased the contrast ratios of header text to above 4.5:1 for smaller text and above 3:1 for larger text.

This makes the text pass level AA of the [Web Content Accessiility Guidelines](http://www.w3.org/TR/WCAG20/).


jbarrett:
Thanks, @tagawa!

tagawa:
Thanks for the quick merge. Will try to check for other accessibility issues.

jbarrett:
Thanks @tagawa. It should be noted that the CSS is getting a bit of an overhaul soonish so I wouldn't spend too much time trawling it - could all be thrown out by the time you get to it.

Thanks again.

tagawa:
Ah, that's good to know - thanks.

PR #553, opened 2015-03-02
Optimized PNG images

tagawa:
Total filesize is around 170kb smaller.

For reference, I used Trimage on Linux but other lossless optimizers should give the same output.


jbarrett:
Many thanks for this, @tagawa!

PR #554, opened 2015-03-02
Close phase 1 submit end point

jbarrett:
@zekiel Explicitly checks for campaign name since some of our nice users have been submitting automatically.


zekiel:
thx

PR #556, opened 2015-03-03
Dev Page - no need to reload when changing type or team values

MariagraziaAlastra:
@russellholt 
[Screencast for the type change](http://gfycat.com/BruisedOrneryCobra)
[Screencast for producer name change](http://gfycat.com/PeacefulThirdCougar) - here the difference with the live version is that if you change a team name from yours to some other user (or if you remove it), the "assign to me" button is shown again without having to reload.


russellholt:
I love it,  way better! :+1:

PR #558, opened 2015-03-03
Dev Pipeline - add filters on the dev view

MariagraziaAlastra:
@russellholt @jagtalon if you're an admin you can choose to see all IAs in development of which you're the producer, or the designer, or the developer.
A regular user (doesn't matter if he has permissions or not) can see all the IAs which he's developing. 
The filtering is done via CSS classes, like we do in the IA Pages Index.

[Screencast](http://gfycat.com/ColorfulAnguishedChameleon) - on my instance TestOne, which is an admin, is producer of "Another IA in development", designer of "Test" and developer of none.

Screenshot (checkbox is unchecked, so we display all the IAs in development):
schermata 2015-03-03 alle 20 55 29


russellholt:
LGTM, go for it :cactus:

MariagraziaAlastra:
Thank you :cactus:

PR #559, opened 2015-03-04
Dev Page - switch from editable to static view without logging out

MariagraziaAlastra:
@russellholt @jdorweiler 
[Screencast](http://gfycat.com/WarmheartedAcceptableCony)
schermata 2015-03-04 alle 21 35 52


jdorweiler:
Thanks!  Looks great :+1:

PR #564, opened 2015-03-09
Fix some deployment warnings

jbarrett:
- `[PkgVersion] no blank line for $VERSION after package DDGC::Web::Controller::InstantAnswer statement on line 1`
- `SQL::Translator::Parser::DBIx::Class::parse(): SQLT attribute 'on_delete' was supplied for relationship 'Topic/instant_answer_topics', which does not appear to be a foreign constraint. If you are sure that SQLT must generate a constraint for this relationship, add 'is_foreign_key_constraint => 1' to the attributes.`

### DIFF
--- a/lib/DDGC/DB/Result/Topic.pm
+++ b/lib/DDGC/DB/Result/Topic.pm
@@ -30,7 +30,7 @@ unique_column name => {
         is_nullable => 0,
 };
 
-has_many 'instant_answer_topics', 'DDGC::DB::Result::InstantAnswer::Topics', 'topics_id', {on_delete => 'cascade'};
+has_many 'instant_answer_topics', 'DDGC::DB::Result::InstantAnswer::Topics', 'topics_id';
 many_to_many 'instant_answers', 'instant_answer_topics', 'instant_answer';
 
 no Moose;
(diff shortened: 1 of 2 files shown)


jbarrett:
@jdorweiler @MariagraziaAlastra Please confirm the schema change leaves the intent of the IA Topic relationships intact. Thanks!

lib/DDGC/DB/Result/Topic.pm
```diff
@@ -30,7 +30,7 @@ unique_column name => {
         is_nullable => 0,
 };
 
-has_many 'instant_answer_topics', 'DDGC::DB::Result::InstantAnswer::Topics', 'topics_id', {on_delete => 'cascade'};
+has_many 'instant_answer_topics', 'DDGC::DB::Result::InstantAnswer::Topics', 'topics_id';
```
MariagraziaAlastra: @jbarrett why did you have to remove `{on_delete => 'cascade'}`? I mean, I understand you got that error shown, but I don't fully understand why

MariagraziaAlastra:
@jbarrett Looks like everything works fine!

jbarrett:
I think they are only disabled in `belongs_to` by default so only need to be specified there:

https://metacpan.org/pod/DBIx::Class::Relationship#belongs_to

MariagraziaAlastra:
Oh, ok. Thanks!

PR #565, opened 2015-03-10
IA Pages - add 'answerbar' column to the InstantAnswer table and display the field on the Dev Page

MariagraziaAlastra:
@russellholt @bsstoner @tommytommytommy @nilnilnil 
Please let me know if I got this right: the column 'answerbar' is a JSON string which, for now, only contains fallback_timeout, but it can easily contain any other value you might need.

You should be able to set the fallback_value for any Spice not yet live inside the Planning panel of the dev page (screenshots below):
schermata 2015-03-10 alle 21 26 28
schermata 2015-03-10 alle 21 26 33

JSON data returned:
schermata 2015-03-10 alle 22 05 43

### DIFF
--- a/lib/DDGC/DB/Result/InstantAnswer.pm
+++ b/lib/DDGC/DB/Result/InstantAnswer.pm
@@ -28,6 +28,13 @@ column description => {
 	is_nullable => 1,
 };
 
+# JSON string cointaining parameters such as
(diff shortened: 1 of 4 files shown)


MariagraziaAlastra:
Merging

nilnilnil:
:+1:

nilnilnil:
Thanks for making ms

PR #566, opened 2015-03-12
IA Pages - move the 'New IA Page' button from the Index to the Dev Pipeline

MariagraziaAlastra:
@russellholt 
schermata 2015-03-12 alle 17 26 57
schermata 2015-03-12 alle 17 27 00


jdorweiler:
Thanks! @MariagraziaAlastra

PR #568, opened 2015-03-12
Dev Page - make description editable

MariagraziaAlastra:
@russellholt a 'metafields' Handlebars template is now appended to the dev page before the dev milestone panels. 
Right now it only contains the description, but its purpose is more generally to contain all the meta fields, aka fields which don't necessarily belong to a specific dev milestone, such as tab name and topics.
schermata 2015-03-12 alle 20 21 15

### DIFF
--- a/src/ia/css/ia.css
+++ b/src/ia/css/ia.css
@@ -1362,3 +1362,7 @@ body.texture {
     -moz-border-radius: 3px;
     -o-border-radius: 3px;
 }
+
+.metafield {
+    margin-bottom: 1em;
+}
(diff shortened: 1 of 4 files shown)


russellholt:
Can we make the description input a lot longer? Most of the fields are much too small in general, but will can on the layout change to address the rest of them.

MariagraziaAlastra:
I'll make it a textarea, thanks!

MariagraziaAlastra:
@russellholt  
schermata 2015-03-12 alle 21 58 00

russellholt:
:+1:  thanks
LGTM

PR #569, opened 2015-03-12
Dev Page - move topic and tab name to metafields.handlebars

MariagraziaAlastra:
@russellholt 
schermata 2015-03-12 alle 23 11 54
schermata 2015-03-12 alle 23 12 52


russellholt:
great :+1:

PR #572, opened 2015-03-13
Dev Page - add allowed topics after appending metafields.handlebars

MariagraziaAlastra:
@russellholt Fixes empty dropdowns


MariagraziaAlastra:
Merging, urgent.

PR #576, opened 2015-03-16
IA Pages - single page - change status to dev milestone

MariagraziaAlastra:
@russellholt looks exactly as before, only that now the value in there is the dev milestone instead of the status
schermata 2015-03-16 alle 17 43 48


MariagraziaAlastra:
Merging, minor

russellholt:
:+1:

PR #577, opened 2015-03-16
IA Pages - add new fields to the editable page

MariagraziaAlastra:
@russellholt first bunch of new fields for the editable page (need to add some other fields in other PRs)
schermata 2015-03-16 alle 22 40 22
schermata 2015-03-16 alle 22 39 08


MariagraziaAlastra:
Merging...

PR #587, opened 2015-03-19
updates table

jdorweiler:
Adding an updates table.  Still in progress
cc: @MariagraziaAlastra @russellholt @jbarrett


jdorweiler:
This is ready for some testing.  Everything seems to work fine to me but I may have missed something.
 
New stuff:
 - Updates are stored in instant_answer_updates
 - TO_JSON method for InstantAnswer::Result.
  - This handles converting json data from the database into json for you.  `is_json` is set in the column attribute for each field that you want to convert.  
  -  also handles returning specific fields for the json endpoint.  see `for_endpt` in column 
- lots of changes to support the update table

Double check the way that the developer updates work.  I'm not sure if the transition from developer in the pipeline to contributor to the live page is correct.

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -307,12 +293,9 @@ sub ia_base :Chained('base') :PathPart('view') :CaptureArgs(1) {  # /ia/view/cal
             $can_edit = 1;
 
             if ($is_admin) {
-                my $edits = get_edits($c->d, $answer_id);
+                my @edits = get_all_edits($c->d, $c->stash->{ia}->id);
```
MariagraziaAlastra: Why not use `$answer_id` instead of `$c->stash->{ia}->id`? Anyway when I'm logged in as admin and I try to see a single IA Page, and thus this line is executed, I get a DB error. I think it's in the get_all_edits() function...

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -937,32 +681,29 @@ sub remove_edit {
...
+    # delete all entries from updates with field
+    my $edits = $d->rs('InstantAnswer::Updates')->search({ instant_answer_id => $ia->id, field => $field });
+    $edits->delete();
 }
 
+# given the IA name return a result set of all edits
+sub get_all_edits {
+    my ($d, $id) = @_;
+    warn "Getting edits for $id" if debug;
+    my @edits = $d->rs('InstantAnswer::Updates')->search( {instant_answer_id => $id} );
```
MariagraziaAlastra: ...This line is probably causing the DB error, because it says "column me.instant_answer_id does not exist"

sql/1.027/updates_table.sql
```diff
@@ -0,0 +1,12 @@
...
+CREATE TABLE "instant_answer_updates" (
+      "id" text NOT NULL,
+      "field" text NOT NULL,
+      "value" text NOT NULL,
+      "timestamp" text NOT NULL,
+      PRIMARY KEY ("id")
+);
+
+ALTER TABLE "instant_answer_updates" ADD CONSTRAINT "instant_answer_updates_fk_id" FOREIGN KEY ("id")
+  REFERENCES "instant_answer" ("id") ON DELETE CASCADE ON UPDATE CASCADE DEFERRABLE;
```
MariagraziaAlastra: @jbarrett is this line right? Because on my instance it doesn't recognize instant_answer_id as a valid column...

jbarrett: Yeah, looks like this DDL for this is wrong:

https://github.com/duckduckgo/community-platform/blob/jd/updates_table/lib/DDGC/DB/Result/InstantAnswer/Updates.pm#L32

It should probably have a primary key also, even if that's also `instant_answer_id`

Running `ddgc_schema_diff` gives me:

```sql
CREATE TABLE "instant_answer_updates" (
  "instant_answer_id" text NOT NULL,
  "field" text NOT NULL,
  "value" text NOT NULL,
  "timestamp" text NOT NULL
);

ALTER TABLE "instant_answer_updates" ADD CONSTRAINT "instant_answer_updates_fk_instant_answer_id" FOREIGN KEY ("instant_answer_id")
  REFERENCES "instant_answer" ("id") ON DELETE CASCADE ON UPDATE CASCADE DEFERRABLE;
```

MariagraziaAlastra:
Thanks @jdorweiler, great job on this! Going to merge and release on monday :+1:

MariagraziaAlastra:
Oh, right: please @jdorweiler remember to remove the updates column in the .sql file! Thanks!

jdorweiler:
@MariagraziaAlastra done.  You will need to run this manually since you already ran the update table sql `ALTER TABLE instant_answer DROP COLUMN updates;`

MariagraziaAlastra:
@jdorweiler I think you need to remove it from InstantAnswer.pm too: https://github.com/duckduckgo/community-platform/blob/jd/updates_table/lib/DDGC/DB/Result/InstantAnswer.pm#L225

MariagraziaAlastra:
Tested (a lot), LGTM, really excellent job here @jdorweiler, this is smart code. Thank you! :cactus:

PR #588, opened 2015-03-19
Dev IA Pages - use the author column from instant_answer_issues to fill the developer column if empty

MariagraziaAlastra:
@russellholt 
schermata 2015-03-19 alle 17 26 42
schermata 2015-03-19 alle 17 26 39


russellholt:
Good idea

PR #599, opened 2015-03-24
Email cleanup

jbarrett:
Cleaning up email modelling - storing it within data blob precludes certain other changes, e.g. selecting only users with email set for notification mail.


jbarrett:
@zekiel  Marked as WIP until some testing on view tomorrow. Thanks!

zekiel:
registration send --> wfm
pw reset email to that confirmed address --> wfm 
no password reset email to unconfirmed address --> wfm 

didn't get it when i shouldn't. did when in should've :) 

are you wrapping the registration message into this as well?

jbarrett:
Many thanks for that, @zekiel - updated texts rolled into this, will probably go out in today's release.

PR #600, opened 2015-03-24
IA Pages - editable page - add new fields

MariagraziaAlastra:
WIP, still missing a bit of functionality there


MariagraziaAlastra:
@russellholt ready
24 - Fathead:
schermata 2015-03-25 alle 18 25 50
Amazon - Spice:
schermata 2015-03-25 alle 18 26 49

russellholt:
LGTM!

russellholt:
only one weird issue in edit mode. checkboxes can be changed before edit is enabled, in which case they change with edit is enabled.

PR #602, opened 2015-03-26
IA Pages - add api_status_page column and make it editable both on the dev page and on the editable live page

MariagraziaAlastra:
@russellholt @moollaza 
Dev page:
schermata 2015-03-26 alle 14 17 26

Editable page:
schermata 2015-03-26 alle 14 19 29

Commit page:
schermata 2015-03-26 alle 14 23 49


russellholt:
:+1:

PR #603, opened 2015-03-27
Dev IA Page - get more info and prevent panels from collapsing when saving

MariagraziaAlastra:
@russellholt @moollaza This is the base for better UX on the dev page. For now the immediate outcome is no collapsing panels, and that if the value wasn't saved it will instantly get back to displaying the last saved one.

[Screencast](http://gfycat.com/ThisMilkyJunebug)


src/ia/js/IAPage.js
```diff
@@ -563,23 +573,18 @@
...
+                            if (data.result && data.result.saved) {
                                 if (field === "dev_milestone") {
                                     location.reload();
                                 } else {
                                     ia_data.live[field] = (is_json && data.result[field])? $.parseJSON(data.result[field]) : data.result[field];
+                                    readonly_templates[panel + "_content"] = Handlebars.templates[panel + "_content"](ia_data);
 
-                                    if (field === "repo" || field === "producer"
-                                        || field === "designer" || field === "developer") {
+                                    var $panel_body = $("#" + panel + " .dev_milestone-container__body");
```
russellholt: this is better

MariagraziaAlastra: :+1:

PR #607, opened 2015-03-31
on /repo return only live

russellholt:
late night fix
may not be the best way, but consider this a placeholder
maybe also include ready ? 

we were including metadata in the repo release for in dev IAs that don't have perl modules, causing internal issues that were addressed separately

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -88,7 +88,11 @@ sub iarepo_json :Chained('iarepo') :PathPart('json') :Args(0) {
 
     my %iah;
 
+IA:
     for my $ia (@x) {
+
+        next IA unless $ia->dev_milestone eq 'live'; # or ready?
+
         my @topics = map { $_->name} $ia->topics;
 
         $iah{$ia->id} = {


lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -88,7 +88,11 @@ sub iarepo_json :Chained('iarepo') :PathPart('json') :Args(0) {
 
     my %iah;
 
+IA:
     for my $ia (@x) {
+
+        next IA unless $ia->dev_milestone eq 'live'; # or ready?
```
jbarrett: Why not select the dev milestone you need?

```perl
my @x = $c->d->rs('InstantAnswer')->search({
    repo => $repo,
    dev_milestone => 'live',
});
```

Also, would HashrefInflator + serialising to JSON directly work for this function?

russellholt: I'm fine with that. I just did it fast at 2:30 am to do something 
See asana

MariagraziaAlastra:
If this is urgent we can merge and then make a new PR with the HashRefInflator?

MariagraziaAlastra:
LGTM

PR #610, opened 2015-03-31
IA Pages - allow removing all topics

MariagraziaAlastra:
@russellholt Right now if you remove all topics from a live IA and then go to the commit page it doesn't display anything; this PR fixes that.


russellholt:
:+1:

PR #611, opened 2015-03-31
IA Pages - remove just the selected value instead of the whole element for topics

MariagraziaAlastra:
@russellholt ...So that the select elements for topics are always three. Also removed a bit of unused code, merging right away


russellholt:
:+1:

PR #613, opened 2015-04-01
WIP: IAPage: Add button for generating screenshots.

jagtalon:
This adds a button in the IA page that lets people take screenshots of the IA.


jagtalon:
Initial demo: https://vid.me/qQrA

jagtalon:
Another demo but on the live page: https://vid.me/MEpg

jagtalon:
Still have to add:
- Notify the user if the user is not available.
- Only show the button to admins.

jagtalon:
@MariagraziaAlastra Is there a good way to check for permissions in screens.handlebars? I see that it gets passed `latest_edits_data` and I don't think the permissions are in there.

MariagraziaAlastra:
@jagtalon quick hack: for the dev page check the existence of the "mark as complete" button; for the live page check the existence of the commit button.

jagtalon:
@MariagraziaAlastra the `screens` template is now using `ia_data` so that it has access to permissions:

screen shot 2015-04-02 at 11 18 03 am
screen shot 2015-04-02 at 11 18 11 am

MariagraziaAlastra:
@jagtalon Great! I like the video previews too :cactus:

jagtalon:
@MariagraziaAlastra I guess only thing left is getting this to work on duck.co which I'm still trying to figure out. `Plack::App::Proxy` should only run on dev servers so we need an Nginx proxy for the screenshot service.

@jbarrett worth adding `Plack::App::Proxy` to the deps?

jagtalon:
@MariagraziaAlastra you can test by running `script/Command.sh` (and also installing Plack::App::Proxy)

src/ia/js/IAPage.js
```diff
@@ -167,25 +167,58 @@
...
+                        $.post("/screenshot/save/" + ia_data.live.id, function(data) {
+                            // Hide the save button.
+                            $(".save-screenshot--button").addClass("hide");
+                            // Put in the image from S3.
+                            $(".ia-single--image-container img").attr("src", "https://images.duckduckgo.com/iu/?u=" + encodeURIComponent(data.screenshots.index));
+                        });
+                    });
+
+                    // Generate a screenshot when the button is clicked.
+                    $(".generate-screenshot--button").click(function() {
```
MariagraziaAlastra: Please change these event handlers to be like in https://github.com/duckduckgo/community-platform/blob/jag/screenshot-button/src/ia/js/IAPage.js#L272

aka `$("body").on("click", ".generate-screenshot--button", function(evt) { ... });`

Since the Handlebars is appended after the page is loaded this will prevent buggy event handling

src/ia/js/IAPage.js
```diff
@@ -167,25 +167,58 @@
...
+                                $(".ia-single--screenshots").show();
+                                $(".ia-single--right").removeClass("dashed-border");
+                            } else {
+                                $(".generate-screenshot--button").html("Generate Screenshot");
+                            }
+                        }).error(function() {
+                            // Revert button to its original state since the server is down.
+                            // Add a little delay.
+                            setTimeout(function() {
+                                $(".generate-screenshot--button").html("Generate Screenshot");
```
MariagraziaAlastra: Maybe print a message error above the button, too? "Ooops, something went wrong - try again!" (just an idea)

src/templates/screens.handlebars
```diff
@@ -1,45 +1,54 @@
...
+          <div class="serp-item"></div>
+        </div>
+      {{/eq}}
+      
+    </div>
     </div>
   </div>
-</div>
+
+  {{#if ../permissions.can_edit}}
```
MariagraziaAlastra: This isn't set for the live page though. You need to modify these lines - just remove the dev milestone check and it should be fine: https://github.com/duckduckgo/community-platform/blob/jag/screenshot-button/src/ia/js/IAPage.js#L58-L64

src/ia/js/IAPage.js
```diff
@@ -167,25 +167,58 @@
...
+                            $(".ia-single--image-container img").attr("src", "https://images.duckduckgo.com/iu/?u=" + encodeURIComponent(data.screenshots.index));
+                        });
+                    });
+
+                    // Generate a screenshot when the button is clicked.
+                    $(".generate-screenshot--button").click(function() {
+                        $(".generate-screenshot--button").html("Generating ...");
+
+                        // Send a POST request with the ID of the IA.
+                        $.post("/screenshot/create/" + ia_data.live.id, function(data) {
```
MariagraziaAlastra: Instead of `ia_data.live.id` please use `DDH_iaid`

src/ia/js/IAPage.js
```diff
@@ -167,25 +167,58 @@
...
+                            // Check if the screenshot that we want is available.
+                            // If it isn't there must be something wrong.
+                            if(data.screenshots.index) {
```
MariagraziaAlastra: Is it possible that data and/or data.screenshots isn't defined/is empty? If that's the case better check `if (data && data.screenshots && data.screenshots.index) { ... }`

src/ia/js/IAPage.js
jagtalon: @MariagraziaAlastra isn't it the same thing? just aliased? https://github.com/jquery/jquery/blob/10399ddcf8a239acc27bdec9231b996b178224d3/src/event/alias.js#L12..L16

What buggy things are you seeing?

src/ia/js/IAPage.js
jagtalon: Good idea!

src/templates/screens.handlebars
jagtalon: Good catch

src/ia/js/IAPage.js
MariagraziaAlastra: Not exactly the same - the way you've written it the event is bound to the element, and when the page is loaded that element doesn't exist yet, it's created afterwards after adding the screens handlebars template. So JQuery might not see it.

The way I suggested you to write binds the event to the body, so when you click it will always trigger.

jagtalon: Oh you meant binding it to the body. Cool makes sense!

jagtalon:
@MariagraziaAlastra updated

src/ia/js/IAPage.js
```diff
@@ -55,28 +55,27 @@
...
-                        ia_data.current = {};
-                        ia_data.current[ia_data.live.dev_milestone] = 1;
-                        var future = {};
-                        var milestone_idx = $.inArray(ia_data.live.dev_milestone, page.dev_milestones_order);
-                        if (milestone_idx !== -1) {
-                            milestone_idx++;
-                            for (var i = milestone_idx; i < page.dev_milestones_order.length; i++) {
-                                future[page.dev_milestones_order[i]] = 1;
-                            }
+                    }
```
MariagraziaAlastra: Sorry for explaning myself poorly, I meant just remove the check for this permissions part. The code afterwards, which sets the current and future dev milestone, isn't needed for live pages, so could you put the check back for lines https://github.com/duckduckgo/community-platform/blob/jag/screenshot-button/src/ia/js/IAPage.js#L66-L77 ?

src/ia/js/IAPage.js
```diff
@@ -167,25 +166,58 @@
...
+                            // Hide the faux-window if the screenshot doesn't exist.
+                            $(".ia-single--screenshots").hide();
+                            $(".ia-single--right").addClass("dashed-border");
 
                             page.imgHide = true;
                         }
                     });
+
+                    // Saves the screenshot to S3.
+                    $("body").on("click", ".save-screenshot--button", function() {
```
MariagraziaAlastra: `function(evt)` don't leave it without arguments

src/ia/js/IAPage.js
```diff
@@ -167,25 +166,58 @@
...
+                        $.post("/screenshot/save/" + DDH_iaid, function(data) {
+                            // Hide the save button.
+                            $(".save-screenshot--button").addClass("hide");
+                            // Put in the image from S3.
+                            $(".ia-single--image-container img").attr("src", "https://images.duckduckgo.com/iu/?u=" + encodeURIComponent(data.screenshots.index));
+                        });
+                    });
+
+                    // Generate a screenshot when the button is clicked.
+                    $("body").on("click", ".generate-screenshot--button", function() {
```
MariagraziaAlastra: Same as above

src/ia/js/IAPage.js
jagtalon: Yup can do!

jagtalon:
Some notifications and failing cases:

- Server is down: https://vid.me/ynFc and https://vid.me/hNE6
- Example query is missing from the metadata: https://vid.me/ijn5
- IA did not show up in the screenshot: https://vid.me/IMJV
- When everything goes well: https://vid.me/YVSs

I'm sure there are a lot more failing cases that I haven't covered.

MariagraziaAlastra:
@jagtalon Excellent job with those error messages!

src/ia/js/IAPage.js
```diff
@@ -73,10 +75,10 @@
                                 future[page.dev_milestones_order[i]] = 1;
                             }
                         }
-
-                        ia_data.future = future;
                     }
 
+                    ia_data.future = future;
```
MariagraziaAlastra: This goes inside the if

src/ia/js/IAPage.js
```diff
@@ -167,25 +169,67 @@
...
+                    // Generate a screenshot when the button is clicked.
+                    $("body").on("click", ".generate-screenshot--button", function(evt) {
+                        var $button =  $(".generate-screenshot--button");
+                        $button.text("Generating ...");
+
+                        // Send a POST request with the ID of the IA.
+                        $.post("/screenshot/create/" + DDH_iaid, function(data) {
+                            // Check if the screenshot that we want is available.
+                            // If it isn't there must be something wrong.
+                            if(data && data.status === "ok" && data.screenshots && data.screenshots.index) {
```
MariagraziaAlastra: :+1:

src/ia/js/IAPage.js
```diff
@@ -167,25 +169,67 @@
...
+                                $button.text("Generate Screenshot");
+                                $(".save-screenshot--button").removeClass("hide");
+
+                                // Show preview image.
+                                $(".ia-single--image-container img").attr("src", data.screenshots.index);
+                                $(".ia-single--screenshots").show();
+                                $(".ia-single--right").removeClass("dashed-border");
+                            } else {
+                                $button.text("Generate Screenshot");
+                                setNotification(data.status);
```
MariagraziaAlastra: Good one

src/ia/js/IAPage.js
jagtalon: ah good catch

jagtalon:
Switching from "Editable" to "Static" is giving me problems at the moment--it's not showing the generate button when there's no screenshot. Not yet sure why. Here's how it looks: https://vid.me/cjRr

src/ia/js/IAPage.js
```diff
@@ -742,6 +804,7 @@
                 }
 
                 if (!this.imgHide) {
+                    $(".ia-single--right").append(templates.screens);
                     $(".ia-single--right").append(templates.live.screens);
```
MariagraziaAlastra: Please remove this line

jagtalon: thanks didn't notice that

jagtalon:
@MariagraziaAlastra update: most of the stuff is ready except for the default screenshot in the IA Page when you're logged out. Here's how it looks: https://vid.me/Llde

MariagraziaAlastra:
@jagtalon Nice!

jagtalon:
@MariagraziaAlastra here's the default screenshot showing when an admin is logged out and there's no screenshot: https://vid.me/yotc

screen shot 2015-04-09 at 2 51 05 pm

jbarrett:
Hey, sorry, missed the `Plack::App::Proxy` question. I think we're past that point, but FWIW I think dev dependencies should get added to cpanfile.

jagtalon:
@MariagraziaAlastra It's ready!!!

screen shot 2015-04-21 at 12 45 00 pm

01n4njx

MariagraziaAlastra:
Awww yiss! Thanks @jagtalon!

jagtalon:
Aw yiss!

PR #614, opened 2015-04-01
Commit Page - don't select any version by default and disable commit button until a version for each field is selected

MariagraziaAlastra:
@moollaza @russellholt 
schermata 2015-04-01 alle 19 27 16
schermata 2015-04-01 alle 19 27 19
schermata 2015-04-01 alle 19 27 22


moollaza:
:+1:

PR #619, opened 2015-04-02
Don't skip username include if userpage headline is set.

jbarrett:
Pass headline to this include instead.

Thanks to @KNTRO for digging this up.


KNTRO:
Thank YOU.

PR #622, opened 2015-04-02
Dev Pipeline - add view for deprecated IAs

MariagraziaAlastra:
@russellholt @moollaza I thought it would be a good idea to group IAs by repo there:
schermata 2015-04-02 alle 21 29 49

Also, since in my understanding a deprecated IA is one which was live and has been taken down at some point, I decided to use the live static page layout, even though the breadcrumb points to the dev pipeline (notice the red label for the deprecated dev milestone):
schermata 2015-04-02 alle 21 29 39


jagtalon:
Works pretty well! I like that we can revive it again.

MariagraziaAlastra:
:cactus:

russellholt:
cool
I wonder how many there will be

and what happens to the code. Removed from the repo I would hope. The file links should probably point to the last release on github before it was live rather than just master

PR #624, opened 2015-04-06
IA Pages - editable page - allow values removal

MariagraziaAlastra:
@russellholt @jdorweiler 
schermata 2015-04-06 alle 17 25 55


jdorweiler:
:+1:

PR #626, opened 2015-04-07
Email uploader when a new PO file is uploaded.

jbarrett:
Same as duckpan upload, have poupload mail a link + the filename for review to the uploader.


jbarrett:
I've added this mailer since, while we have automated po uploading for some texts, there's no notification for new POs for review.

PR #627, opened 2015-04-07
Dev IA Pages - refresh panels height on autocommit to prevent content from overflowing

MariagraziaAlastra:
@russellholt previously the content overflowed when setting the type to "fathead"


src/ia/js/IAPage.js
```diff
@@ -655,6 +651,17 @@
             'ready'
         ],
 
+        setMaxHeight: function($obj_set) {
```
russellholt: better

PR #628, opened 2015-04-07
set unsafe column to zero

jdorweiler:
@MariagraziaAlastra


MariagraziaAlastra:
Thanks @jdorweiler, LGTM :+1:

PR #630, opened 2015-04-08
Microservice arch with blog POC

jbarrett:
@MariagraziaAlastra @nilnilnil @zekiel @jdorweiler @russellholt 

This is currently Not Working, but I'll go a little into what's different here from the last PR and what's needed to get moving.

Biggest change from the last discussed PR is using the plugin architecture to wrap DDGC::Config and PSGI Sessions (so e.g. just `use DDGC::Web::Plugin::Session` should populate a user variable in the request, if there is a current user)

Re-using the model and templates is possibly too messy - attempting a more significant change here has exposed some stuff which won't work or is impractical, e.g.

- Instances of `DDGC` are heavily relied on by **most or all** components (usually passed in on instantiation)
- `DDGC::DB` does not use a standard schema constructor / connect call so can't just be used as a DBIx::Class schema (with Plugin::DBIC say)
- Templates rely heavily on functionality in [`DDGC`'s Xslate constructor](https://github.com/duckduckgo/community-platform/blob/72a6e3930ca59fee4d087aa966a0c715b79c329f/lib/DDGC.pm#L233-L450) or [`DDGC::Web::View::Xslate`](https://github.com/duckduckgo/community-platform/blob/72a6e3930ca59fee4d087aa966a0c715b79c329f/lib/DDGC/Web/View/Xslate.pm)

So for this to work without depending too much on the extant architecture (i.e. to be as discrete as possible from the start), there needs to be parallel development on a DB Schema and views which implement parts that makes sense and leaves behind parts that don't.

There are modules like [`DDGC::Markup`](https://github.com/duckduckgo/community-platform/blob/72a6e3930ca59fee4d087aa966a0c715b79c329f/lib/DDGC/Markup.pm) which are instantiated as a member of, and passed a reference to, a `DDGC` instance. These types of instances should probably be plain modules with exports. Where a dependency on session or request parameters exists, they can be passed as parameters, though efforts should be made to reduce this type of dependency anyway.

No insurmountable challenges yet :)

More tomorrow...


lib/DDGC/Web/App/Base.pm
```diff
@@ -0,0 +1,33 @@
...
+            DDGC::Web::Plugin::Config
+            DDGC::Web::Plugin::Session
+          /
+        ) {
+            $_->import::into($caller);
+        }
+    }
+};
+
+1;
```
jbarrett: Re `package DDGC::Web::App::Base`

`Import::Into` performs an import step on any module of your choosing (so `'Foo'->import::into('Bar')` effectively inserts `use Foo` into `package Bar`).

This makes `DDGC::Web::App::Base` a "base module" (**not** a base class) which does the setup, exports and everything else `use`ing those modules would.

The reason for this is, each app or service has a number of required modules to do config and setup. While you can copypaste from a working application, this is prone to error. I wanted to make creating a new app much simpler. ([See this diff](https://github.com/duckduckgo/community-platform/commit/d9a547f))

@MariagraziaAlastra @nilnilnil @jdorweiler @russellholt Any thoughts? Neat or horrific? I'm not sure.

russellholt: Not horrific! Cool

jdorweiler: horrifically cool :+1:

jbarrett: Just to note : This pattern does result in a deprecation warning (from Session plugin calling the keyword registered by the DBIC plugin), but that is on its way to being resolved elsewhere.

https://github.com/ironcamel/Dancer2-Plugin-DBIC/pull/9

nilnilnil:
Can you expound a bit on why there is so much overlap between service and app?

jbarrett:
@nilnilnil Yeah, since they access a chunk of the same resources they have a lot of modules in common. We could have a `Web::Base` or something which does the common `import::into`s?

What I was going for was to be able to have a single `use` in a new App and you have everything you need. I suppose two `use`s is almost as good :)

We could also have `Web::Base` do `import::into(caller(2))`, keeping a single use, but this strikes me as messy and prone to error.

nilnilnil:
Is DDG::Web a role there? Move up?

jbarrett:
`DDGC::Web` is basically the existing Catalyst app. I'd like to keep this as-is for servicing the extant architecture.

I experimented with Moo roles for composing Apps, but it didn't really work with this framework - it's applications are "just" modules, not classes.

I should perhaps move the new parts out of the `DDGC::Web` space altogether to distinguish them properly.

nilnilnil:
*shrug* I'm cool with this direction until a better one leaps off the page from a DRI perspective.

jbarrett:
I'm with you @nilnilnil - I can at least move the common includes out into their own module.

That `caller(2)` business was a brain fart on my part - not required.

jbarrett:
With `DDGC::Web::Service::Blog` mounted on `/blog.json`:

```
$ curl http://127.0.0.1:5003/blog.json/
{"posts":[{"content":"This is a content","user":"1","title":"Test One 31","id":59,"uri":"test-one-31"},{"content":"Other This is a content","user":"1","title":"Other Test One 30","id":58,"uri":"my-other-test-one-30"},{"content":"This is a content","user":"1","title":"Test One 30","id":57,"uri":"test-one-30"},{"content":"This is a content","user":"1","title":"Test One 29","id":55,"uri":"test-one-29"} ....
```

There are a number of ways to select pages / posts-per-page:

```
$ curl http://127.0.0.1:5003/blog.json/page/3
{"posts":[{"content":"Other This is a content","user":"1","title":"Other Test One 4","id":6,"uri":"my-other-test-one-4"},{"content":"This is a content","user":"1","title":"Test One 4","id":5,"uri":"test-one-4"},....

$ curl http://127.0.0.1:5003/blog.json/page/3/pagesize/3
{"posts":[{"content":"This is a content","user":"1","title":"Test One 27","id":51,"uri":"test-one-27"},{"content":"Other This is a content","user":"1","title":"Other Test One 26","id":50,"uri":"my-other-test-one-26"},{"content":"This is a content","user":"1","title":"Test One 26","id":49,"uri":"test-one-26"}]}

$ curl 'http://127.0.0.1:5003/blog.json/?page=2&pagesize=3'
{"posts":[{"content":"This is a content","user":"1","title":"Test One 29","id":55,"uri":"test-one-29"},{"content":"Other This is a content","user":"1","title":"Other Test One 28","id":54,"uri":"my-other-test-one-28"},{"content":"This is a content","user":"1","title":"Test One 28","id":53,"uri":"test-one-28"}]}
```

Next steps are:

- DDGC-agnostic Markup renderer
- Comments integration
- POST handler, integrated with forms in `/my`
- Key is ID, not URL - this is madness. URL can be a "slug", but requiring uniqueness can go.
- More stuff as it comes...

jbarrett:
I just don't like the look of the source in [`Dancer2::Plugin::DDGC::Request`](https://github.com/duckduckgo/community-platform/blob/john/blog-service/lib/Dancer2/Plugin/DDGC/Request.pm) but for now, it is what it is.

Anyway, it provides `ddgcr_get()` and `ddgcr_post()` which I guess I'll go into with some examples.

In the `/blog` rendering app, we decide we want the default set of posts from `/blog.json`:

```perl
var 'posts' => ddgcr_get '//blog.json';
```

- `//` denotes that we want to request a path off root.
  - Use '/' for paths in the current app
  - This **looks** like URI semantics, but it isn't
- Returns a `HTTP::Response` instance containing our Perl data (`$res->{ddgcr}`)

We want page 2, and now we're only rendering 10 posts per page:

```perl
var 'posts' => ddgcr_get '//blog.json', { page => 2, pagesize => 10 };
```

There's a new blog post (new endpoint doesn't currently exist):

```perl
ddgcr_post '//blog.json/new', $json_string;
```

...or...

```perl
ddgcr_post '//blog.json/new', { title => 'Awesome update', content => '...' };
```

There's no pressing need for this POST route to be private, since the Plack session id is passed along in these requests where it exists - user authorisation can be performed as normal.

russellholt:
:+1: looking good so far

jbarrett:
`->for_user` allows us to have `current_user` context available in ResultSets.

An example might be if you want to guard against a service handing out content for a moderators forum to users without access:

```perl
rset('Comment')->for_user( var 'user' )->forum($post);
```

Then in the Comment RS you can do something like (this doesn't make total sense, just illustrative):

```perl
sub forum {
    my ( $self, $post_id ) = @_;
    return $self->search_rs(\['1=0'])
        if !( $self->current_user->can_access_forum_post( $post_id ) );
    return $self->search_rs( {
        context => 'forum',
        context_id => $forum_id,
    } );
}
```

I was going to have `->for_user` take a user ID or User instance as parameter, but making the schema available to the `coerce` clause was getting messy for little gain.

Another option would be to wrap `rset` so `->for_user` is always done, but I'm not sure that doesn't seem like some invisible magic.

jbarrett:
Thinking on this, the rs should really throw when the user has no access, with service catching and returning 403 or such. Anyway, added a `->none` should you wish to "nullify" a query plan.

It should also be stated, that if you need `current_user` for your plan, the call should be first in the chain.

jbarrett:
While `var 'user'` will give us the current user if it's in the request, useful behaviours for user role and no logged in user need to be defined. `Dancer2::Plugin::DDGC::UserRole` is an attempt to do this for you.

For example, we want to guard against non-admin users POSTing to `/blog.json/new` - creating new blog posts.

```perl
post '/new' => user_is 'admin' => sub {
    # rset stuff goes here
};
```

If there is no user logged in or the user is not an admin, this will return a HTTP 403.

If you just require a logged in user with any role, say `user_is 'user'`

If you want to guard everything under a path, you could do something like:

```perl
post qr{^/translation.*} => user_is 'translation_manager' => sub {
    pass;
};

post '/translation/delete/:id' => sub {
    # rset stuff goes here
};
```

This regex will capture everything going to /translation. `pass` sends the request to the next matching route in the file.

In Apps, the behaviour is a little different. If there is no user, it will redirect to the login page. If there is a user, but without access then HTTP 403 is returned. If we redirect, `last_url` is set in the session so we redirect back to the initial requested route after login. Setting `not_last_url` will skip storing the current request url in the session.

You can set this, or other plugin config options, by passing a hashref with required values to `user_is`:

```perl
get '/account' => user_is 'user', { not_last_url => 1 } => sub {
    template 'account';
};
```

jbarrett:
@MariagraziaAlastra https://metacpan.org/pod/DBIx::Class::Helper::ResultSet::Shortcut::HRI

You might be interested in that - enabling it in the resultset base class allows you to enable the HashRefInflator by adding `->hri` to the query chain. You could insert it here to try it out, I think:

https://github.com/duckduckgo/community-platform/blob/d29525c/lib/DDGC/DB/Base/ResultSet.pm#L8-L14

It's also enabled for the schema in this PR.

jbarrett:
So, what's new...

[A forum / comment markup renderer is taking shape](https://github.com/duckduckgo/community-platform/blob/15b59e4/lib/DDGC/Util/Markup.pm). This is to provide [`DDGC::Markup`](https://github.com/duckduckgo/community-platform/blob/01e0db8/lib/DDGC/Markup.pm) services without the DDGC instance dependency (which is mainly used in `DDGC::Markup` to ascertain if the user would like to open links in a new window - a single boolean).

`ddgcr_get` and `ddgcr_post` are now also part of your DDGC instance (`$c->d->ddgcr_post` or such....) You can also call them as detailed above or like:

```perl
my $res = ddgcr_post [ 'Blog', 'admin', 'post', 'new' ], { $form_data };
```
or 
```perl
my $res = ddgcr_post [ qw/ Blog admin post new / ], { $form_data };
```
This is like requesting `//blog.json/admin/post/new`.

The blog editing part of the user profile manager in `DDGC::Web` posts edits to the blog.json service - this appears to work well. New posts should be simple enough too.

Lots of other internal-consistency type stuff added, let me know if you have questions.

MariagraziaAlastra:
>@MariagraziaAlastra https://metacpan.org/pod/DBIx::Class::Helper::ResultSet::Shortcut::HRI

Thanks @jbarrett!

jbarrett:
Apps are rendering.

Exciting blank page

Exciting login dialog

jbarrett:
OK, `->for_user` is gone. It was on crack.

Result and ResultSet classes still have a `current_user` accessor.

lib/DDGC.pm
```diff
@@ -127,6 +129,47 @@ has uuid => (
...
 
+sub _apply_session_to_req {
+	my ( $c, $req ) = @_;
+	$req->header(
+		Cookie => 'plack_session=' . $c->req->env->{'psgix.session.options'}->{id},
+	);
+}
+sub _ref_to_uri {
+	my ( $route ) = @_;
+	my $service = '/' . lc( shift @{$route} ) . '.json';
```
nilnilnil: are these refs guaranteed to be `'ARRAY'`?

lib/DDGC.pm
```diff
@@ -127,6 +129,47 @@ has uuid => (
...
+	my ( $uri ) = join  '/', @{$route};
+	return "$service/$uri";
+}
+sub ddgcr_get {
+	my ( $self, $c, $route, $params ) = @_;
+	$route = _ref_to_uri( $route  ) if ( ref $route eq 'ARRAY' );
```
nilnilnil: hahaha -- yup :smile:

lib/DDGC/Base/Web/App.pm
```diff
@@ -0,0 +1,29 @@
+package DDGC::Base::Web::App;
+
+# ABSTRACT: Base module for Applications (rendering services)
+
+use Import::Into;
+
+use strict;
+use warnings;
+use utf8;
```
nilnilnil: why? no utf8 in this source?

lib/DDGC/Base/Web/Common.pm
```diff
@@ -0,0 +1,47 @@
...
+            Dancer2::Plugin::Params::HashMultiValue
+            Dancer2::Plugin::DDGC::Config
+            Dancer2::Plugin::DDGC::Session
+            Dancer2::Plugin::DDGC::Request
+            Dancer2::Plugin::DDGC::Validate
+            Dancer2::Plugin::DDGC::UserRole
+            Dancer2::Plugin::DDGC::SchemaApp
+          /
+        ) {
+            $_->import::into($caller);
```
nilnilnil: as mentioned before, this is pretty gangster -- but scary.

lib/DDGC/Config.pm
```diff
@@ -5,7 +5,7 @@ use Moose;
 use File::Path qw( make_path );
 use File::Spec;
 use File::ShareDir::ProjectDistDir;
-use DDGC::Static;
+eval 'use DDGC::Static';
```
nilnilnil: where do you print errors for this?

lib/DDGC/DB/Result/User/Blog.pm
```diff
@@ -1,202 +1,108 @@
 package DDGC::DB::Result::User::Blog;
+
 # ABSTRACT: Result class of blog posts
 
-use Moose;
-use MooseX::NonMoose;
-extends 'DDGC::DB::Base::Result';
+use Moo;
```
nilnilnil: This looks largely cosmetic. Can you expound on what this gets us? It looks like it's removing a level of indirection from what the base class really should be?

lib/DDGC/DB/Result/User/Blog.pm
```diff
@@ -1,202 +1,108 @@
...
+    is_nullable => 0,
 };
 
 column teaser => {
-	data_type => 'text',
-	is_nullable => 0,
+    data_type   => 'text',
+    is_nullable => 0,
 };
-sub html_teaser {
```
nilnilnil: the fuq?

lib/DDGC/DB/Result/User/Blog.pm
```diff
@@ -1,202 +1,108 @@
...
-	data_type => 'timestamp with time zone',
-	set_on_create => 1,
-	set_on_update => 1,
+    data_type     => 'timestamp with time zone',
+    set_on_create => 1,
+    set_on_update => 1,
 };
 
 belongs_to 'user', 'DDGC::DB::Result::User', 'users_id';
-belongs_to 'language', 'DDGC::DB::Result::Language', 'language_id', { join_type => 'left' };
```
nilnilnil: Can you briefly expound on what was going on here? It looks like transaction / triggers on column change. Where is the functionality now?

lib/DDGC/Schema/Result/User.pm
```diff
@@ -0,0 +1,152 @@
...
+
+column flags => {
+    data_type => 'text',
+    is_nullable => 0,
+    serializer_class => 'JSON',
+    default_value => '[]',
+};
+
+# TODO: Migrate 'flags' to 'roles'
+sub is {
```
nilnilnil: I think there could be some confusion of role vs. moo role.

nilnilnil: I don't think that's enough to change this though -- especially since once you read it a bit it becomes pretty clear.

lib/DDGC/Schema/Result/User/Blog.pm
```diff
@@ -0,0 +1,186 @@
...
+
+column updated => {
+    data_type     => 'timestamp with time zone',
+    set_on_create => 1,
+    set_on_update => 1,
+};
+
+belongs_to 'user', 'DDGC::Schema::Result::User', 'users_id';
+
+after insert => sub {
```
nilnilnil: Is this the new code that represents the old trigger update column shit?

lib/DDGC/Schema/ResultSet/User/Blog.pm
```diff
@@ -0,0 +1,55 @@
...
+        {
+            $self->me . 'company_blog' => 1,
+            ( $self->current_user && $self->current_user->is('admin') )
+            ? ()
+            : ( $self->me . 'live' => 1 ),
+        }
+    );
+}
+
+sub util_json_string { JSON::MaybeXS->new->allow_nonref(1)->encode(shift) }
```
nilnilnil: Why is this needed with the `TO_JSON` serializer defined?

nilnilnil:
@jbarrett I've added some questions to continue our dialog. I think this is pretty sweet. I like that you did the testing & documentation too. Even the sub params are nicer and more clear. Please also prioritize understanding performance. If it's out of the application right now, like in the DB, then pull that bit out and return 1 or something.

lib/DDGC/Base/Web/App.pm
jbarrett: No, but this gets applied to source which deals with templates / other data so there's potential for utf8 in source.

lib/DDGC/Config.pm
jbarrett: I don't - don't care.

It's supposed to be the community-platform CSS bundle but it's either under-utilised or over-utilised - either way we don't care about the result of this.

lib/DDGC/DB/Result/User/Blog.pm
jbarrett: This is  basically a temporary table def because notifications still rely on it - I was too hasty moving it, and reinstated a minimal version just for column definitions.

lib/DDGC/DB/Result/User/Blog.pm
jbarrett: It never existed - spurious (or premature) relationship.

row could be `translation_of` other row in `language` - YAGNI. When we do, we can implement it.

lib/DDGC/Schema/Result/User.pm
jbarrett: Hmmm, for this particular function I think 'flags' just happens to be an inaccurate name. We could call it access_level or anything else. Role seemed the simplest thing at the time.

There is a roles column, currently completely unused.

lib/DDGC/Schema/ResultSet/User/Blog.pm
jbarrett: For searching within JSON columns - like topics.

jbarrett: Inherited from DDGC::DB::ResultSet - definitely needs a rethink.

I like postgres arrays. Do you? :)

lib/DDGC/DB/Result/User/Blog.pm
jbarrett: Can't change the names just yet - it's the post fragment that goes in the index.

lib/DDGC/Schema/Result/User/Blog.pm
jbarrett: This would trigger event generation for notifications. Genuinely, I've read the code many times and have no idea what `seen_live` is really supposed to be for.

Shim for existing notification system compat coming (I think this is easy..), `seen_live`probably needs to go away.

jbarrett:
#### New dependencies

**Debian:**

libcurl4-gnutls-dev

**Perl:**

SPARKY/Test-HTTP-Server-0.03.tar.gz
SYP/Net-Curl-0.37.tar.gz
SYP/LWP-Protocol-Net-Curl-0.023.tar.gz
CHISEL/Catalyst-Plugin-Session-PSGI-0.0.2.tar.gz
ETHER/Import-Into-1.002004.tar.gz
INGY/Test-Base-0.88.tar.gz
INGY/YAML-1.15.tar.gz
INGY/Test-YAML-1.05.tar.gz
INGY/Spiffy-0.46.tar.gz
PETDANCE/Test-Memory-Cycle-1.04.tar.gz
XSAWYERX/Dancer2-0.160001.tar.gz
LDS/Devel-Cycle-1.12.tar.gz
ADAMK/Template-Tiny-1.12.tar.gz
MAUKE/Return-MultiLevel-0.04.tar.gz
MAUKE/Data-Munge-0.095.tar.gz
ETHER/Test-Warnings-0.021.tar.gz
RUSSELLJ/Dancer2-Session-PSGI-0.008.tar.gz
MIYAGAWA/Plack-Middleware-Session-0.30.tar.gz
LEONT/Module-Build-Tiny-0.039.tar.gz
KAZEBURO/Cookie-Baker-0.05.tar.gz
SATOH/Test-Time-0.04.tar.gz
SATOH/Test-Name-FromLine-0.13.tar.gz
JBARRETT/Dancer2-Plugin-RootURIFor-0.02.tar.gz
IRONCAMEL/Dancer2-Plugin-DBIC-0.0009.tar.gz
IRONCAMEL/DBICx-Sugar-0.0001.tar.gz
TOBYINK/Test-Modern-0.013.tar.gz
DAGOLDEN/Test-API-0.005.tar.gz
TOBYINK/Exporter-Tiny-0.042.tar.gz
RJBS/Test-Deep-0.115.tar.gz
ILMARI/SQL-Translator-0.11021.tar.gz
HOBBESTIG/Dancer2-Plugin-Feed-1.142060.tar.gz
RSIMOES/Dancer2-Template-Xslate-v0.0.7.tar.gz
BOBTFISH/Text-Markdown-1.000031.tar.gz 
TOKUHIROM/HTML-TreeBuilder-LibXML-0.25.tar.gz
ARISTOTLE/Hash-MultiValue-0.16.tar.gz
ROKR/Hash-Merge-Simple-0.051.tar.gz
ETHER/URI-1.67.tar.gz
PEVANS/Scalar-List-Utils-1.42.tar.gz
DOY/Try-Tiny-0.22.tar.gz
MMCCLENN/HTTP-Validate-0.45.tar.gz
EXODIST/Test-Simple-1.001014.tar.gz
LEONT/Module-Build-0.4212.tar.gz
DAGOLDEN/CPAN-Meta-2.150001.tar.gz
DAGOLDEN/Parse-CPAN-Meta-1.4414.tar.gz
DAGOLDEN/CPAN-Meta-YAML-0.016.tar.gz
MIKO/String-Util-1.24.tar.gz
MIKO/Number-Misc-1.2.tar.gz

jbarrett:
Merge notes:

- cpanfile includes HTML::FormatText::WithLinks
- /ddgc_static is gone (plackup scripts and templates)
- dzil / travis no longer require App::DuckPAN

jbarrett:
@MariagraziaAlastra @nilnilnil 

There's some work needed to merge this, since a variety of architectural changes have gone in (mainly deprecation of DDGC::Static and plack based dev services). I'm happy enough to merge it as-is and fix issues that arise in a small additional PR (well, I expect it to be small). It should take max two days, but I would aim for a couple of hours. So:

- Are we happy to merge this?
- Will my plan to break the dev branch clash with deployment time lines?

MariagraziaAlastra:
@jbarrett The plan for IA Pages was to deploy tomorrow, so for me it kinda depends on the type of issues we could face after merging this: if they're small enough that they can be gone for tomorrow at max 3:00 AM EST then ok, go for it.
Otherwise... we could try to speed up things and deploy today in about six hours.

jbarrett:
@MariagraziaAlastra I think I'll have to hold off.

nilnilnil:
:godmode:

PR #635, opened 2015-04-10
InstantAnswer.pm - get live, ready and qa IAs in the repo JSON endpoint

MariagraziaAlastra:
@russellholt @jagtalon So they can be deployed on test machines when in QA


jagtalon:
@MariagraziaAlastra nice!

russellholt:
Thanks

PR #644, opened 2015-04-16
IA Pages: fix attribution

MariagraziaAlastra:
WIP


MariagraziaAlastra:
Sneak peek:
screenshot-maria duckduckgo com 5000 2015-04-29 16-27-22

MariagraziaAlastra:
[Username check in action](http://gfycat.com/UnluckyColorfulCottonmouth)

jdorweiler:
:+1:

jagtalon:
- It's possible to have the same username in multiple times (this has the same GitHub username and URL):
screen shot 2015-04-30 at 1 07 20 pm
- I think it should call `/ia/usercheck` not only when the `<select>` is changed, but also when the user is done typing. Otherwise usernames that don't exist will get accepted.
screen shot 2015-04-30 at 1 11 22 pm
- The UI elements work on FF but span two lines in Chrome and Safari and Opera:
screen shot 2015-04-30 at 1 15 49 pm
screen shot 2015-04-30 at 1 16 14 pm
screen shot 2015-04-30 at 1 17 19 pm

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -5,6 +5,7 @@ use namespace::autoclean;
 use Try::Tiny;
 use Time::Local;
 use JSON;
+use Net::GitHub::V3;
```
jagtalon: Add this to `dist.ini`?

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -344,7 +345,9 @@ sub ia_json :Chained('ia_base') :PathPart('json') :Args(0) {
                       url => 'https://github.com/'.$pull_request{author}
                   );
 
-                  my $value = to_json \%dev_hash;
+                  my @dev_array = [\%dev_hash];
+
+                  my $value = to_json \@dev_array;
```
jagtalon: What about `my $value = to_json [\%dev_hash]` since this is the only place we use `@dev_array` it seems. Might be cleaner.

MariagraziaAlastra:
@jagtalon 

>- I think it should call `/ia/usercheck` not only when the `<select>` is changed, but also when the user is done typing. Otherwise usernames that don't exist will get accepted.

No they won't, the green/red highlight is just to give an immediate visual feedback to the user, the check will be performed again before saving the edit to the updates table.
Anyway I agree that it might be confusing.

Thanks for the feedback!

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -675,6 +721,26 @@ sub add_edit {
     });    
 }
 
+sub check_github {
+    my ($username) = @_;
+
+    my $result = 0;
+    my $token = $ENV{DDGC_GITHUB_TOKEN} || $ENV{DDG_GITHUB_BASIC_OAUTH_TOKEN};
+    my $gh = Net::GitHub->new(access_token => $token);
```
jagtalon: Will this try to authenticate on GH every time we call `check_github`?

jagtalon:
@MariagraziaAlastra I get a console error when I try to save a user in duck.co that doesn't exist. Without opening the console, I wouldn't have known that that was the case because there's no visual feedback:

screen shot 2015-04-30 at 1 40 08 pm

MariagraziaAlastra:
Yeah that's what I meant

jagtalon:
@MariagraziaAlastra It felt broken because I couldn't click "Done" :)

jagtalon:
@MariagraziaAlastra putting non-ascii characters in the username triggers an exception (which results in a 500 error on the page):

screen shot 2015-04-30 at 1 50 38 pm
screen shot 2015-04-30 at 1 51 16 pm

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -545,6 +577,28 @@ sub save_edit :Chained('base') :PathPart('save') :Args(0) {
     return $c->forward($c->view('JSON'));
 }
 
+sub usercheck :Chained('base') :PathPart('usercheck') :Args() {
+    my ( $self, $c ) = @_;
+
+    my $username = $c->req->params->{username};
+    my $type = $c->req->params->{type};
```
jagtalon: Should probably fall back to `''` just in case there are no params. Accessing the URL gives me `Use of uninitialized value $type in string eq at /home/ubuntu/community-platform/script/../lib/DDGC/Web/Controller/InstantAnswer.pm line 588.`.

```perl
my $username = $c->req->params->{username} || '';
my $type = $c->req->params->{type} || '';
```

would probably be better.

lib/DDGC/Web/Controller/InstantAnswer.pm
MariagraziaAlastra: yes, and it's required, otherwise we would get rate limited

MariagraziaAlastra:
@russellholt 
screenshot-maria duckduckgo com 5000 2015-04-30 21-12-43
screenshot-maria duckduckgo com 5000 2015-04-30 22-21-39
screenshot-maria duckduckgo com 5000 2015-04-30 22-22-14

MariagraziaAlastra:
@russellholt Since it's ready I'll briefly explain the problem and the solution

###The problem

We need reliable attribution data, but so far only the new data in the developer column is really reliable, and the urls associated can only be either duck.co accounts or github accounts; the old data, on the other hand, which is in the attribution column, has a different structure and allows for different types of associated urls, so we can find anything from a twitter account to a personal website.

Example of the current developer column structure:
```json
{
    "name": "Jason Dorweiler",
    "url": "https://github.com/jdorweiler
}
```

Example of the attribution column structure:

```json
{
  "Jason Dorweiler": [
    {"loc": "jdorweiler",
      "type": "github"
    },
    {"loc": "@jdorweiler",
     "type": "twitter"
    }
  ],
  "Jag Talon": [
    {"loc": "jagtalon",
     "type": "github"
    }
  ]
}
```


###The solution

Restructuring the developer column to be an array of hashes, each for a single developer, with each hash containing the name, the url and the type (can be github, duck.co or legacy, for twitter/personal websites), in which will be included the old data from the attribution column; this is done by a script in the sql folder (I put it there because it's meant to run just once).

Example of the new developer column structure:

```json
[
    {"name": "Jason Dorweiler",
     "url": "https://github.com/jdorweiler",
     "type": "github"
    },
   {"name": "Jag Talon",
    "url": "https://github.com/jagtalon",
    "type": "github"
   }
]
```

This means we also need to update the way the developer data is saved and retrieved on the back-end, how it's showed on the front-end, and that we need a new UI/UX for editing it, and it's all included in this PR.

First notable change is: now you can manually add github accounts, too, because there's a check for username validity on the back-end. On the editable page, as showed by the screencast I posted in one of the comments above, you can only edit github and duck.co attribution types - legacy types can't be modified, but you can remove them if you want to.

The first input on the left is for the name - you can leave it blank if you want it to be the same as the username; on the right there's a dropdown for choosing types and an input for the username; e.g. for adding Jason's github account you would select "github" from the dropdown and then type "jdorweiler" in the input; after typing and pressing on the return key, or after changing the type in the dropdown, a check will be performed to see if the user you specified exists: if it does the dropdown background color and the input border color will be green; if it doesn't exist they will turn red.

Anyway the check will be performed again when you attempt to save by clicking on the "done" button, and if one of the entries will have an invalid username the edit won't be saved.

On the dev page you can specify more devs by separating them with commas, as for the perl dependencies and triggers fields; if you're specifying a duck.co user you don't need to add the whole url when typing (even if it will be displayed the whole url after saving), but if you want to add a github account, please paste the whole url, or it will assume it's a complat username and will perform the wrong check.

Also updated the data needed to make the filtering for developers work on the dev pipeline.

###Still to do

Check usernames before they're displayed in the editable page (some users like to change their names) and also on input focusout, better UI/UX for the dev page, make it possible to add more than one developer each time you edit on the live editable page.

###So, after this goes live, is it all fixed?

No. We should manually go through all the IAs and check the developer field. If there's any legacy value we should at least verify that the url does't point to malicious websites, and possibly if it's still valid; we might need to remove entries, edit them... it will be a bunch of work, but it needs to be done.

Please let me know if you have any more questions

russellholt:
Good, thorough writeup. And the current plan is to eventually require duck.co account names only, which will further simplify this.

MariagraziaAlastra:
Thanks! If you want to test it before merging... :)

PR #647, opened 2015-04-21
IA Pages - show meta_id as id in the JSON endpoints and update frontend accordingly

MariagraziaAlastra:
@russellholt 
screenshot-maria duckduckgo com 5000 2015-04-21 18-15-44
screenshot-maria duckduckgo com 5000 2015-04-21 18-45-01
screenshot-maria duckduckgo com 5000 2015-04-21 18-45-24

### DIFF
--- a/lib/DDGC/DB/Result/InstantAnswer.pm
+++ b/lib/DDGC/DB/Result/InstantAnswer.pm
@@ -15,14 +15,14 @@ sub u { [ 'InstantAnswer', 'view', $_[0]->id ] }
 
 column id => {
 	data_type => 'text',
-    for_endpt => 1
 };
 primary_key 'id';
 
 # editable ID
 column meta_id => {
     data_type => 'text',
-    for_endpt => 1
+    for_endpt => 1,
+    alias => 'id'
 };
 
 # userland name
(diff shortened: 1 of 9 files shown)


lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -15,14 +15,14 @@ sub u { [ 'InstantAnswer', 'view', $_[0]->id ] }
...
-    for_endpt => 1
 };
 primary_key 'id';
 
 # editable ID
 column meta_id => {
     data_type => 'text',
-    for_endpt => 1
+    for_endpt => 1,
+    alias => 'id'
```
russellholt: alias?

MariagraziaAlastra: yeah, new custom attribute :) it's used to display meta_id as id in the JSON endpoint which use the TO_JSON method to get the columns, such as ia_json

MariagraziaAlastra: (ia_json --> single page JSON endpoint)

russellholt:
:+1:  let's do it!

PR #651, opened 2015-04-23
Dev Pipeline - make use of custom column attributes, fix meta_id and developer values

MariagraziaAlastra:
WIP


MariagraziaAlastra:
@russellholt ready
Thanks @jdorweiler for the help with TO_JSON!

PR #657, opened 2015-04-24
IA Pages - rename dev milestones

MariagraziaAlastra:
@russellholt as discussed
screenshot-maria duckduckgo com 5000 2015-04-24 20-52-45
screenshot-maria duckduckgo com 5000 2015-04-24 20-56-14


russellholt:
is it necessary to rename them internally?

MariagraziaAlastra:
Yeah of course, it will break otherwise.

PR #659, opened 2015-04-27
Fix for double encoding in blog RSS feed.

jbarrett:
This explicitly disables body encoding, then sets the encoding in the HTTP header.


jooon:
:+1:

PR #660, opened 2015-04-28
Gruntfile.js: Compass now compiles SASS files.

jagtalon:
Adding a .scss file in `src/ia/css` (this directory could be changed to `src/ia/scss` but we can do that in the future) will now get compiled and minified when we run `grunt`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,8 @@ local-lib
 *.log
 *.index
 node_modules/*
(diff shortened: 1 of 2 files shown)


jagtalon:
To @jdorweiler or @MariagraziaAlastra

MariagraziaAlastra:
LGTM :+1:

PR #661, opened 2015-04-30
more fathead metadata

jdorweiler:
Added a few more fields to the fathead source options and switched skip_end to use a text box.

### DIFF
--- a/src/templates/commit_page.handlebars
+++ b/src/templates/commit_page.handlebars
@@ -374,10 +374,10 @@
                             </span>
                         </li>
                         <li>
-                            <i class="section-group__item icon-check{{#is_false original.src_options.source_skip}}-empty{{/is_false}}" data-field="source_skip"/>
-                            <span class="check-txt">
-                                Source skip
-                            </span>
+                            <label class="left">Source skip</label>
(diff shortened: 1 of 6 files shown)


src/templates/commit_page.handlebars
```diff
@@ -368,6 +362,18 @@
                             </span>
                         </li>
                         <li>
+                            <i class="section-group__item icon-check{{#is_false original.src_options.is_fanon}}-empty{{/is_false}}" data-field="is_fanon"/>
+                            <span class="check-txt">
+                                Is Fanon
```
russellholt: what's this mean?

jdorweiler: Sounds like these are mediawiki source that can sometimes have articles on real-world topics.  We want to be careful about showing them.  I didn't pick the name :smile:

russellholt: Cool

MariagraziaAlastra:
LGTM!

src/templates/commit_page.handlebars
mrshu: @russellholt Sorry for the late reply, @jdorweiler did describe what it really is. Pretty much every bigger wikia has that.

PR #664, opened 2015-05-04
IAPage.js - editable page - add unlimited inputs instead of just one for each edit attempt

MariagraziaAlastra:
@russellholt now it's possible to add multiple items to fields such as triggers, perl dependencies and developer instead of just one at a time


MariagraziaAlastra:
screenshot-maria duckduckgo com 5000 2015-05-04 18-32-56

russellholt:
:+1:

PR #665, opened 2015-05-04
Dev IA Pages - make name editable

MariagraziaAlastra:
@russellholt @moollaza 
screenshot-maria duckduckgo com 5000 2015-05-04 20-18-28


russellholt:
:+1:

PR #685, opened 2015-05-20
IA Pages - add is_stackexchange field and make src_domain editable for longtails

MariagraziaAlastra:
@russellholt @jdorweiler 
screenshot-maria duckduckgo com 5000 2015-05-20 17-44-03
screenshot-maria duckduckgo com 5000 2015-05-20 17-44-48

### DIFF
--- a/lib/DDGC/DB/Result/InstantAnswer.pm
+++ b/lib/DDGC/DB/Result/InstantAnswer.pm
@@ -347,6 +347,12 @@ column src_id => {
     for_endpt => 1
 };
 
+column is_stackexchange => {
+    data_type => 'integer',
+    is_nullable => 1,
+    for_endpt => 1
+};
+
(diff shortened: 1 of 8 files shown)


jdorweiler:
:+1:  thanks!

lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -347,6 +347,12 @@ column src_id => {
     for_endpt => 1
 };
 
+column is_stackexchange => {
```
russellholt: why are such things independent columns?

MariagraziaAlastra: Because it's an independent field, so we treat it as independent column. At least I think it is independent, from my understanding :)

PR #690, opened 2015-05-21
Reorganize frontend assets:

jagtalon:
We can now use SCSS with DDGC. The structure has also been organized:

```
src/
     js/ia
     js/ddgc
     scss/ia
     scss/ddgc
```

This is in preparation for merging https://github.com/duckduckgo/community-platform-static as well. It should sit in `src/ddgc/static`, I think.


MariagraziaAlastra:
:+1:

PR #692, opened 2015-05-21
Merge of duckduckgo/community-platform-static into duckduckgo/community-platform

jagtalon:
It looks like it works, but needs more testing and we also have to remove instances of `DDGC::Static`. It's currently pulling in the SCSS files that we need in https://github.com/duckduckgo/community-platform/blob/1ceeacb5949eedce7f2b7430dc76cfe87987ca4f/src/scss/content/main.scss and we're compiling it as `content.scss`.

So currently `ddgc.css`, `ia.css`, and `content.css` is being built, but we can easily merge them into just one file in the future.

For @MariagraziaAlastra and @sdougbrown.


jagtalon:
@jbarrett could you check it out as well? Is it possible to deploy it in view without releasing or maybe only releasing locally?

src/scss/content/_global.scss
```diff
@@ -0,0 +1,989 @@
...
+				background: -moz-linear-gradient(top,  #ffffff 0%, #f0f0f0 100%);
+				background: -webkit-gradient(linear, left top, left bottom, color-stop(0%,#ffffff), color-stop(100%,#f0f0f0));
+				background: -webkit-linear-gradient(top,  #ffffff 0%,#f0f0f0 100%);
+				background: -o-linear-gradient(top,  #ffffff 0%,#f0f0f0 100%);
+				background: -ms-linear-gradient(top,  #ffffff 0%,#f0f0f0 100%);
+				background: linear-gradient(to bottom,  #ffffff 0%,#f0f0f0 100%);
+				-pie-background: linear-gradient(to bottom,  #ffffff 0%,#f0f0f0 100%);
```
russellholt: what the heck

jbarrett:
@jagtalon :+1: I am away this afternoon for the weekend. Cool if I evaluate on Monday?

You can do a `dzil build` and deploy the tarball to view with:

`script/ddgc_release.sh view DDGC-1.037.tar.gz`

(As long as your key is installed for ddgc@view)

...though deployments take place on Friday, which means a lot of stuff goes on in view so check out the community-platform room first

MariagraziaAlastra:
@jbarrett Monday is perfect, I didn't want to roll this out today anyway. 

@jagtalon Thanks, that's a lot of work! I'm delaying merging this just to make sure everything is alright :smile:

jbarrett:
One thing to check is 404s in firebug's network output and fonts load properly in new forum post and feedback pages :)

jagtalon:
@jbarrett @MariagraziaAlastra thanks! Yes I'm definitely cool with that--I don't wanna break the site. I've been checking the console and haven't found anything yet.

script/ddgc_dev_server.psgi
```diff
@@ -22,7 +22,6 @@ builder {
 
     mount '/' => DDGC::Web->new->psgi_app;
 
-    mount "/ddgc_static" => Plack::App::File->new( root => DDGC::Static::sharedir )->to_app;
```
jbarrett: We should also lose the `use DDGC::Static` above

jbarrett:
So, I ran `grunt` to generate static files and `script/ddgc_dev_server.sh` to kick off a new dev instance. Is this correct?

I am getting 404s for the files ddg-feedback.ttf and ddg-feedback.woff, meaning there are no icons on feedback pages:

missing feedback icons

I also get no icons on the front page or site buttons (though I don't think they use the feedback font):

missing frontpage icons

Thanks!

jagtalon:
@jbarrett off, alright i'll check it out!

jbarrett:
@jagtalon Thanks. Let me know what help I can provide.

script/ddgc_dev_server.psgi
jagtalon: @jbarrett thanks good find!

jagtalon:
@jbarrett @MariagraziaAlastra fonts should be working now. Will try to deploy on a staging server.

jbarrett:
:+1: Staging LGTM, thanks @jagtalon!

MariagraziaAlastra:
Thanks @jagtalon! :cactus:

PR #696, opened 2015-05-27
updated text

zekiel:
//cc @jbarrett not sure if this will render correctly -- can you verify on your box before releasing?


zekiel:
:+1:

PR #697, opened 2015-05-27
Gruntfile.js: Modifying .handlebars with grunt watch doesn't work.

jagtalon:
This will actually make `grunt watch` compile the Handlebars files.


jagtalon:
For @jdorweiler or @MariagraziaAlastra

PR #698, opened 2015-05-27
Screenshot: Disable button if there is no test machine or not live.

jagtalon:
Here's a video: https://vid.me/eixe

### DIFF
--- a/src/js/ia/IAPage.js
+++ b/src/js/ia/IAPage.js
@@ -253,22 +253,7 @@
                             $generate_screenshot.addClass("generate-screenshot--disabled");
                         }
                     }
-
                     enableScreenshotButton();
-                    $("body").on("change", ".test_machine", function(evt) {
-                        if($(this).val()) {
-                            test_machine = $(".test_machine option[value='" +  $(this).val() + "']").text();
-                        } else {
-                            test_machine = null;
-                        }
-
-                        enableScreenshotButton();
-                    });
-
-                    $("body").on("change", ".example_query", function(evt) {
-                        example_query = $(this).val();
-                        enableScreenshotButton();
-                    });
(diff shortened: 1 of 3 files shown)


jagtalon:
For @MariagraziaAlastra

MariagraziaAlastra:
Hey @jagtalon! It should become active when both the test machine and the example query are in place - if we have the test machine, but no example query, I guess we won't be able to generate a screenshot anyway :)

jagtalon:
@MariagraziaAlastra ah good point!

jagtalon:
@MariagraziaAlastra done. Screencast: https://vid.me/w3Hs (slow internet connection, though)

src/js/ia/IAPage.js
```diff
@@ -240,9 +240,43 @@
...
+                        var $generate_screenshot = $(".generate-screenshot--button");
+                        if((test_machine && example_query) || ia_data.live.dev_milestone === "live") {
+                            $generate_screenshot.removeClass("generate-screenshot--disabled");
+                        } else {
+                            $generate_screenshot.addClass("generate-screenshot--disabled");
+                        }
+                    }
+
+                    enableScreenshotButton();
+                    $("body").on("change", ".test_machine", function(evt) {
```
MariagraziaAlastra: There's no need to create two new events for calling the enableScreenshotButton function: you can check inside the autocommit function (somewhere [here](https://github.com/duckduckgo/community-platform/blob/jag/disable-screenshot/src/js/ia/IAPage.js#L752)) if the saved field was either the test machine or the example query, and in that case, call the function.

MariagraziaAlastra:
Thanks @jagtalon! :+1:

PR #701, opened 2015-05-28
Admin option to unsub a user from all notification

jbarrett:
@zekiel


zekiel:
nice -- good addition.

PR #723, opened 2015-06-05
IA Dev Page - make editing the developer field easier

MariagraziaAlastra:
@jagtalon And now make it shine :)

### DIFF
--- a/root/static/css/content.css
+++ b/root/static/css/content.css
@@ -1,3 +1,4 @@
+@charset "UTF-8";
 /* line 1, ../../src/scss/content/_reset.scss */
 html, body, div, span, object, iframe,
 h1, h2, h3, h4, h5, h6, p, blockquote, pre,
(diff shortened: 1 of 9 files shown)


jagtalon:
@MariagraziaAlastra maybe it shouldn't close when the input is wrong.

https://vid.me/Sg5W

MariagraziaAlastra:
It doesn't close, it's just that it gets refreshed after the attempted saving, and in this case it didn't save anything, so that's why it looks like that :)

jagtalon:
@MariagraziaAlastra what if it didn't try to save when the input is wrong and just changed color?

MariagraziaAlastra:
Yeah let me try that, thanks!

jagtalon:
@MariagraziaAlastra similarly, when the first input is ok, and the second input is bad, they both go away: https://vid.me/Tcqf.

Thanks!

MariagraziaAlastra:
That's weird, I'll look into that, thank you!

MariagraziaAlastra:
@jagtalon Fixed, but when the input is invalid the button is only partially red:
screenshot-maria duckduckgo com 5000 2015-06-05 16-18-06

MariagraziaAlastra:
Alright, thanks @jagtalon! :+1:

PR #724, opened 2015-06-05
add pr field and dev_milestone to end point

jdorweiler:
Added dev_milestone to the end point so we can check the status of the IA before trying to run tests on it. 
The pr field seem to be missing too.  


lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -382,6 +383,12 @@ column forum_link => {
     is_nullable => 1
 };
 
+column pr => {
```
MariagraziaAlastra: @jdorweiler Aren't we using the issues table for PRs?

MariagraziaAlastra:
Thanks @jdorweiler! :+1:

PR #731, opened 2015-06-08
use to_json and from_json

jdorweiler:
Fixes the messed up encoding in the updates table. 
selection_078


MariagraziaAlastra:
:+1:

PR #732, opened 2015-06-08
skip pr field when rebuilding

jdorweiler:
The database update script fails on the pr field so just skip it.  You can get these from running ghIssues if you really need them.


MariagraziaAlastra:
:+1:

PR #733, opened 2015-06-08
added ddg attribution

jdorweiler:
Add developer info to any IA that doesn't already have it.


MariagraziaAlastra:
:+1:

PR #739, opened 2015-06-10
Activity feed

jbarrett:
Examples:

Default format is markdown. Let's generate some activity.

Where something is:

```
$ curl -H 'Content-Type: application/json' -X POST -d '{"secret":"asdfasdf", "type":"where", "description":"The index is [here](/)"}' http://127.0.0.1:5002/activityfeed.json/new 2>/dev/null | json_xs
{
   "ok" : 1,
   "status" : 200,
   "result" : {
      "created" : 1433931616,
      "raw_description" : "The index is [here](/)",
      "id" : 21,
      "type" : "where",
      "description" : "<p>The index is <a href=\"/\">here</a></p>"
   }
}
```

Where an IA is:

```
$ curl -H 'Content-Type: application/json' -X POST -d '{"secret":"asdfasdf", "type":"where_ia", "description":"The airlines IA is [here](https://duck.co/ia/view/airlines)"}' http://127.0.0.1:5002/activityfeed.json/new 2>/dev/null | json_xs
{
   "ok" : 1,
   "status" : 200,
   "result" : {
      "created" : 1433931648,
      "raw_description" : "The airlines IA is [here](https://duck.co/ia/view/airlines)",
      "id" : 22,
      "type" : "where_ia",
      "description" : "<p>The airlines IA is <a href=\"https://duck.co/ia/view/airlines\">here</a></p>"
   }
}
```

An IA has gone live:

```
$ curl -H 'Content-Type: application/json' -X POST -d '{"secret":"asdfasdf", "type":"live", "description":"[Arqade is LIVE](https://duckduckgo.com/?q=defcon+strategies&ia=qa)"}' http://127.0.0.1:5002/activityfeed.json/new 2>/dev/null | json_xs
{
   "ok" : 1,
   "status" : 200,
   "result" : {
      "created" : 1433931672,
      "raw_description" : "[Arqade is LIVE](https://duckduckgo.com/?q=defcon+strategies&ia=qa)",
      "id" : 23,
      "type" : "live",
      "description" : "<p><a href=\"https://duckduckgo.com/?q=defcon+strategies&amp;ia=qa\">Arqade is LIVE</a></p>"
   }
}
```

All activity:

```
$ curl http://127.0.0.1:5002/activityfeed.json 2>/dev/null | json_xs
{
   "ok" : 1,
   "status" : 200,
   "activity" : [
      {
         "created" : 1433931672,
         "raw_description" : "[Arqade is LIVE](https://duckduckgo.com/?q=defcon+strategies&ia=qa)",
         "id" : 23,
         "type" : "live",
         "description" : "<p><a href=\"https://duckduckgo.com/?q=defcon+strategies&amp;ia=qa\">Arqade is LIVE</a></p>"
      },
      {
         "created" : 1433931648,
         "raw_description" : "The airlines IA is [here](https://duck.co/ia/view/airlines)",
         "id" : 22,
         "type" : "where_ia",
         "description" : "<p>The airlines IA is <a href=\"https://duck.co/ia/view/airlines\">here</a></p>"
      },
      {
         "created" : 1433931616,
         "raw_description" : "The index is [here](/)",
         "id" : 21,
         "type" : "where",
         "description" : "<p>The index is <a href=\"/\">here</a></p>"
      }
   ]
}
```

Filtering:

```
$ curl http://127.0.0.1:5002/activityfeed.json?filter=where 2>/dev/null | json_xs
{
   "ok" : 1,
   "status" : 200,
   "activity" : [
      {
         "created" : 1433931616,
         "raw_description" : "The index is [here](/)",
         "id" : 21,
         "type" : "where",
         "description" : "<p>The index is <a href=\"/\">here</a></p>"
      }
   ]
}
```

```
$ curl http://127.0.0.1:5002/activityfeed.json?filter=where* 2>/dev/null | json_xs
{
   "ok" : 1,
   "status" : 200,
   "activity" : [
      {
         "created" : 1433931648,
         "raw_description" : "The airlines IA is [here](https://duck.co/ia/view/airlines)",
         "id" : 22,
         "type" : "where_ia",
         "description" : "<p>The airlines IA is <a href=\"https://duck.co/ia/view/airlines\">here</a></p>"
      },
      {
         "created" : 1433931616,
         "raw_description" : "The index is [here](/)",
         "id" : 21,
         "type" : "where",
         "description" : "<p>The index is <a href=\"/\">here</a></p>"
      }
   ]
}
```

In lieu of API Keys (which we probably should have for this sort of thing) there is a shared secret used from the environment.


jbarrett:
```
$ curl http://127.0.0.1:5003/activityfeed.json?filter=iapage*grasshoppers 2>/dev/null | json_xs
{
   "ok" : 1,
   "status" : 200,
   "activity" : [
      {
         "created" : 1436436747,
         "raw_description" : "Instant Answer Page updated : [Grasshoppers](/ia/view/grasshoppers)",
         "id" : 34,
         "type" : "iapage_update_grasshoppers",
         "description" : "<p>Instant Answer Page updated : <a href=\"/ia/view/grasshoppers\">Grasshoppers</a></p>"
      },
      {
         "created" : 1436436738,
         "raw_description" : "New Instant Answer Page created : [Grasshoppers](/ia/view/grasshoppers)",
         "id" : 33,
         "type" : "iapage_new_grasshoppers",
         "description" : "<p>New Instant Answer Page created : <a href=\"/ia/view/grasshoppers\">Grasshoppers</a></p>"
      }
   ]
}
```
```
$ curl http://127.0.0.1:5003/activityfeed.json?filter=iapage_new* 2>/dev/null | json_xs
{
   "ok" : 1,
   "status" : 200,
   "activity" : [
      {
         "created" : 1436436738,
         "raw_description" : "New Instant Answer Page created : [Grasshoppers](/ia/view/grasshoppers)",
         "id" : 33,
         "type" : "iapage_new_grasshoppers",
         "description" : "<p>New Instant Answer Page created : <a href=\"/ia/view/grasshoppers\">Grasshoppers</a></p>"
      },
      {
         "created" : 1436436710,
         "raw_description" : "New Instant Answer Page created : [Locusts](/ia/view/locusts)",
         "id" : 32,
         "type" : "iapage_new_locusts",
         "description" : "<p>New Instant Answer Page created : <a href=\"/ia/view/locusts\">Locusts</a></p>"
      },
      {
         "created" : 1436436638,
         "raw_description" : "New Instant Answer Page created : [Crickets](/ia/view/crickets)",
         "id" : 29,
         "type" : "iapage_new_crickets",
         "description" : "<p>New Instant Answer Page created : <a href=\"/ia/view/crickets\">Crickets</a></p>"
      }
   ]
}
```

sql/1.0xx/activity_feed.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+CREATE TABLE "activity_feed" (
+  "id" serial NOT NULL,
+  "created" timestamp with time zone NOT NULL,
+  "type" text NOT NULL,
```
kablamo: This is really simple which is awesome.  Another idea which is more complex is to use tags.  More flexible than hierarchical stuff and less string manipulation than this solution.

How to build a schema with tags (in case you are like me and didn't know off the top of your head):
http://skm.zoomquiet.io/data/20060423142114/index.html
http://stackoverflow.com/questions/20856/recommended-sql-database-design-for-tags-or-tagging

sql/1.0xx/activity_feed.sql
```diff
@@ -0,0 +1,13 @@
...
+  "description" text NOT NULL,
+  "format" text DEFAULT 'markdown' NOT NULL,
+  "for_user" bigint,
```
kablamo: so if we send a msg to every user and we have 300 users, we'll have to add 300 rows to this table.  is that ok?  depends how many users we get i guess.  and how many notifications.  and if we ever delete stuff from this table.

kablamo: Do we need a 'sent' column to indicate the notification has been emailed...?

jbarrett: If `for_user` is null, it's for everyone. I've started working on a `for_role` where you can notify classes of users (i.e. admins, community leaders, translation managers...). If we have to add 300 columns for a single notification I think it's broken.

I don't think we need a 'sent' column - if the mailer job runs hourly we should select stuff which happened in the last hour (rounded down to $hour:00:00).

jbarrett: The idea of adding the same notification for a handful of users is OK, I think (e.g. mailing stats to 3 or 4 people).

The idea of `for_user` is to target stuff like replies,@ -mentions....

sql/1.0xx/activity_feed.sql
jbarrett: That might work better - I was thinking of places this `type` filtering broke down and it's pretty easy to find them. Let's say we focus on areas of interest, e.g. "I want to know when any IA related to Pokemon has status changed to 'live'." it breaks down pretty quickly but this is inarguably a useful kind of filter.

The string manipulation also smells, I agree completely.

I was also considering a secondary meta-stash which is a postgres array of strings (again, free-form) so the subscription could look something like:
```perl
{
    type => 'iapage%updated',
    meta => [ 'status:live', 'category:pokemon', ],
}
```

I think having *n* tags is ultimately simpler and just as flexible. Applications can designate the hierarchy themselves, `tag1` is their top-level category, `tag2` is action, `tag3` name, `tag4` what's new, tag5 categories...

```perl
{
    tag1 => 'ia',
    tag2 => 'update',
    tag3 => 'pokemon',
    tag4 => 'status:live',
    tag5 => 'reference,geek,special-interest'
}
```

There's still a bit of string-y stuff in there, but 'status:live' can be an exact string match (it's an invented convention, but not as smelly as hierarchical-wildcarding :smile:).

Hmmm, pokemon isn't even an IA category - pretend I contrived a real world example *cough*. Going to give it some thought.

Thanks!

jbarrett:
@kablamo What kind of email will you need to send? I was thinking of the content over the weekend - currently we are catering for single lines of text where you click through to duck.co for the content. There are a number of ways to extend this, a 'content' column, a 'URI' column (somewhere you can retrieve the data as JSON or something required)... Current elements use a `context_object` which is a related resultset name + row id.

kablamo:
I feel like I'm missing context -- I don't really understand.  Why do we need a content column?  Also the last sentence, I don't understand.

I was just going to email a sentence or two of text.  It might be a short paragraph.  With a link.  I could keep it to a sentence if I had to I suppose.

jbarrett:
OK, I think that works just fine in this model.

In the current notifications email there is support for emailing, say, comment content inline. This is achieved by linking the notifications themselves to entities - this is the context object. It is made up of a DBIC Result(Set) name and ID. This is the role which does this:

https://github.com/duckduckgo/community-platform/blob/65e8a80/lib/DDGC/DB/Role/HasContext.pm

This is also used to link comments back to their parent object (forum thread, blog post, translation token...)

Happy to forget this for now.

lib/DDGC/Schema/Result/ActivityFeed.pm
```diff
@@ -0,0 +1,68 @@
...
+has renderer => (
+    is => 'ro',
+    lazy => 1,
+    builder => '_build_renderer',
+);
+sub _build_renderer {
+    DDGC::Util::Markup->new;
+}
+
+sub render_description{
```
kablamo: This is cool

jbarrett: Thanks.

[DDGC::Util::Markup](https://github.com/duckduckgo/community-platform/blob/0dc490e/lib/DDGC/Util/Markup.pm) is intended to replace [DDGC::Markup](https://github.com/duckduckgo/community-platform/blob/0dc490e/lib/DDGC/Markup.pm) - I think it's about 90% ready - little bit of clean up and some rendering quirks to fix.

jbarrett:
@kablamo I've been giving this some thought while working on other problems this week. While we can do stuff like adjacency lists to build hierarchical data structures in the db, they wind up being non-trivial to join or categorise at `SELECT` time... I think the flat set of numbered tags ends up being the simplest and most flexible thing.

Everything done but the naming!

And the relationships, subscriptions, mailer... Oh.

jbarrett:
Designing subscriptions now... and there are multiple potential approaches including but not limited to:

- An object which stores named [SQL::Abstract](https://metacpan.org/pod/SQL::Abstract) stanzas (hashrefs) to build DBIC query plans.
  - User subs are then a table of user, named sub entries
  - Two db ops:
    - Select subscription stanza names (+inflate to stanzas from subscription types instance)
    - Use stanzas to build Activity query plan

This starts to break down pretty quickly when free-form text stuff turns up in subscriptions (e.g. all activity related to 'Pokemon' IA) since stanzas need to be flexible AND the additional info needs to be denoted in the subscription table.

Let's say our subscription types object includes something like the following:

```perl
sub ia_all_activity {
    my ( $self, $ia ) = @_
    +{
        category = >'ia',
        meta1 => $ia,
    }
}
```

...our "subscription" needs to store users_id, 'ia_all_acitivity' and any other potentially-free-form additional information (`$ia` here, possibly serialised perl data structures elsewhere). The uniqueness and intent of a row (is user 123 following all activity on Pokemon IAs?) becomes more difficult to control (adding / removing subscriptions).

Another approach is to have the `user_subscriptions` table have the same columns that describe an `activity_feed` entry, that is: category, action, meta1-3.

In this model, we still have the subscription types object with SQL::Abstract stanzas, but they are used to manage user subscriptions rather than build an activity feed query plan. With this approach we can join directly from user -> activity (I think).

I also have this notion, possibly flawed (think: accuracy, referential integrity, performance / indexing, in-band control messaging / text munging), of using `LIKE` for meta* columns instead of straight equality. This would allow some amount of flexibility:

Say we decide to use meta2 as a category field, delimited (or wrapped) by some character, say '**:**', just because... then a subscription to a category can look like:

```perl
+{
    category => 'ia',
    action => undef,
    meta1 => undef,
    meta2 => '%:special-interest:%',
    meta3 => undef,
}
```

We have this delimiter because we don't want a subscription to 'car' to match on the category 'cards'. Can this be made reliable? Does `LIKE` give you the heebie-jeebies? :)

lib/DDGC/Schema/Result/User/Subscription.pm
```diff
@@ -0,0 +1,89 @@
...
+    is_nullable => 1,
+};
+
+belongs_to 'user', 'DDGC::Schema::Result::User', 'users_id';
+
+has_many 'activity', 'DDGC::Schema::Result::ActivityFeed', sub {
+    my ( $args ) = @_;
+
+    my $activity = $args->{foreign_alias};
+    my $subscription = $args->{self_alias};
```
kablamo: Nice. didn't know about these aliases

jbarrett:
Hoping c57c1a1 explains my intent here. Via subscriptions we can join activity on:

- category
- action (if defined)
- meta1-3 (like, if defined)
- users_id (if for_user defined in activity)
- role (if for_role defined in activity)

lib/DDGC/Schema/Result/User/Subscription.pm
```diff
@@ -0,0 +1,89 @@
...
+};
+
+column meta3 => {
+    data_type => 'text',
+    is_nullable => 1,
+};
+
+belongs_to 'user', 'DDGC::Schema::Result::User', 'users_id';
+
+has_many 'activity', 'DDGC::Schema::Result::ActivityFeed', sub {
```
kablamo: should be activities right?

kablamo: Oh i get it.  nevermind.

lib/DDGC/Schema/Result/User/Subscription.pm
```diff
@@ -0,0 +1,89 @@
...
+                "$activity.meta3" => { 'like' => "$subscription.meta3" },
+            ],
+            -or => [
+                "$activity.for_user" => { '=' => undef },
+                "$activity.for_user" => { '=' => "$subscription.users_id" },
+            ],
+            -or => [
+                "$activity.for_role" => { '=' => undef },
+                "$activity.for_role" => {
+                    -in => \"SELECT role FROM user_role WHERE users_id = $subscription.users_id",
```
jbarrett: So this really, really looks like something that should be a bound parameter prepared statement, right? A glance at this here interpolation / literal-expansion-in-SQL-stringery should rightly make you scream.

...except `$subscription` is a table alias and not some arbitrary value. I'm pretty sure this is, if not The Right Thing, fairly safe.

lib/DDGC/Schema/Result/User/Subscription.pm
jbarrett: Yeah, the definition is a bit bonked, but thinking `$user->subscription->activity` is ultimately more readable.

lib/DDGC/Web/Service/ActivityFeed.pm
```diff
@@ -0,0 +1,86 @@
...
+    rset('ActivityFeed')->search_rs({
+        ( $params->{filter} )
+            ? ( type => { -like => $params->{filter} =~ s/\*/%/gr } )
+            : (),
+        for_user => { '=' => [ uniq( undef, $params->{user} ) ] }
+    }, {
+        order_by => { -desc => 'me.id' },
+        rows     => $params->{pagesize},
+        page     => $params->{page},
+    });
```
kablamo: Arguably more readable this way:
```
rset('ActivityFeed')->search_rs
    ->type(-like => $filter)
    ->for_user([ undef, $params->{user} ])
    ->order_by( -desc => 'me.id' )
    ->limit($params->{pagesize})
    ->page($params->{page});
```
https://metacpan.org/pod/DBIx::Class::Helper::ResultSet::Shortcut
you would have to build your own `type()` sub in `ActivityFeed::ResultSet`.  

Just an idea.

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -805,6 +805,12 @@ sub save_edit :Chained('base') :PathPart('save') :Args(0) {
         }
     }
 
+    $c->ddgcr_post( [ 'ActivityFeed', 'new' ], {
```
kablamo: Cool.  This is a wip so all my comments are unfair.  But I'd say there should probably be a more user friendly way of doing this.

jbarrett:
The query when there is no user (hence no subscriptions) should be:

```perl
{
    for_user => { '=' => undef },
    for_role => { '=' => undef },
}
```

i.e. just return all other activity as a "public" view.

lib/DDGC/Web/Service/ActivityFeed.pm
```diff
@@ -0,0 +1,86 @@
...
+    my $user = var 'user';
+    $params->{for_user} = $user->id if $user;
+    +{
+        activity => [
+            activity_page($params)->all,
+        ]
+    };
+};
+
+post '/new' => sub {
```
kablamo: super minor thing: why not just post to `/` instead of `/new`?  I just thought `/activityfeed.json/new` was weird.

lib/DDGC/Web/Controller/InstantAnswer.pm
kablamo: Also I think the ui needs to work from a cronjob.  I'm thinking:
```
my $client = DDGC::ActivityFeed::Client->new;
$client->post(
    type => $type,
    description => $description,
    meta1 => $meta1,
    ...
);
my @updates = $client->get(...);
```

lib/DDGC/Web/Service/ActivityFeed.pm
kablamo: fwiw, mst gave me a recommendation that I never pass params to `search_rs()`.  Instead he always creates a sub in the ResultSet class (like I indicated for `type()` above).   its hard to explain, but i have found that advice really really works for me.  it is more typing though.  but all the query logic is in the ResultSet class and that is really useful in a couple ways.

jbarrett: That's what the relationship I defined yesterday was for (`type` is gone, replaced by `category`, `meta*` etc), so if we have a user we can say:

```perl
$user->subscriptions->activity
```

Basically, this query is out of date.

Agree on pushing searches to subs in the rs, but I am also very lazy and like having the query plan defined at the point of contact... and this is why we have fat in all layers. Yeah, think the plans-in-rs approach is probably better for this project.

I import a few specific shortcuts in the rs base class, no reason not to pull in the full set. Thanks!

jbarrett: Apropos, popped up in the timeline JUST now:

https://twitter.com/OvidPerl/status/624140719968555008 :)

lib/DDGC/Web/Service/ActivityFeed.pm
jbarrett: Going for "explicit naming" but not sure if it's hitting the mark.

lib/DDGC/Web/Controller/InstantAnswer.pm
jbarrett: Yeah, having this in the controller makes no sense anyway, since we can do Moo* hooks in Result classes to generate activity (same way notifications work now).

Cron jobs can also `use DDGC::Web::Base::Common;` and have access to the same facilities as other applications, including `rset`.

[`ddgcr_get/post`](https://github.com/duckduckgo/community-platform/blob/c2a62a2/lib/Dancer2/Plugin/DDGC/Request.pm) themselves could probably use improvement. If we have to abstract the abstraction, the abstraction is broken. Luckily there's not too much invested in it.

In this instance, anything in DDGC::* should (ab)use the Schema so these calls are going away.

lib/DDGC/Web/Service/ActivityFeed.pm
jbarrett: Get routes should probably be '/all' and '/mine'

kablamo: so that would be `/activityfeed.json/new` right?  i thought that was a weird url.  but its no big deal.  its fine.

lib/DDGC/Web/Service/ActivityFeed.pm
kablamo: thats an interesting tweet.  i wonder how far to take it.  should you not use a result class in a controller?  i feel like thats taking it too far.  but it depends.

lib/DDGC/Schema/Result/User/Subscription.pm
kablamo: it doesn't bother me :)

lib/DDGC/Web/Service/ActivityFeed.pm
jbarrett: Wondering what's weird about it.

lib/DDGC/Web/Service/ActivityFeed.pm
jbarrett: You need **some** handle on the data, since you are responsible for informing the template engine... but yeah, it's pretty obviously a breach of separation-of-concerns to have large chunks of query plan hanging about what should be a thin request router.

So the request router has the parameters which inform the query plan, but I don't think it can get much simpler than something like:

```perl
 $rs->for_user                        
    ->order_by( { -desc => 'me.id' } )
    ->rows( $params->{pagesize} )     
    ->page( $params->{page} )         
```

Is the alternative to pass all those in as parameters to a single resultset method? In order to be immediately readable it would need to be a hashref of params anyway, and just as chunky (if not more so).

I think chaining rs methods like the above in the controller is far more explicit / readable and an acceptable level of other-layer-access.

MVC, so simples!

https://duckduckgo.com/?q=mvc&ia=images

sql/1.0xx/00_activity_feed.sql
```diff
@@ -0,0 +1,33 @@
...
+
+ALTER TABLE "user_subscription" ADD CONSTRAINT "user_subscription_fk_users_id" FOREIGN KEY ("users_id")
+  REFERENCES "users" ("id") DEFERRABLE;
+
+
+CREATE TABLE "activity_feed" (
+  "id" serial NOT NULL,
+  "created" timestamp with time zone NOT NULL,
+  "category" text NOT NULL,
+  "action" text,
```
nilnilnil: Action should be required as well?

jbarrett: :+1: I think it should be for `activity_feed` - thanks!

lib/DDGC/Web/Service/ActivityFeed.pm
```diff
@@ -0,0 +1,103 @@
...
+                ? { for_user => $params->{for_user} }
+                : (),
+        });
+    } catch {
+        $error = 1;
+    };
+
+    bailout( 500, "Something went wrong" ) if ( $error );
+
+    use DDP; p $result;
```
nilnilnil: :loudspeaker:

nilnilnil:
tests?

lib/DDGC/Schema/Result/User/Subscription.pm
```diff
@@ -0,0 +1,89 @@
...
+                }
+            ]
+        }
+    };
+
+};
```
jbarrett: Time to teach computers to feel pain, my [SQL::Abstract](https://metacpan.org/pod/SQL::Abstract)-fu failing me miserably here:

```sql
SELECT "me"."id", "me"."created", "me"."category", "me"."action", "me"."meta1", "me"."meta2", "me"."meta3", "me"."description", "me"."format", "me"."for_user", "me"."for_role" 
  FROM "user_subscription" "user_subscription__row" 
  JOIN "activity_feed" "me" 
    ON ( "me"."for_role" IS NULL OR "me"."for_role" IN ( 
        SELECT role 
          FROM user_role 
        WHERE users_id = user_subscription__row.users_id
       ) ) AND "me"."category" = 'user_subscription__row.category' 
WHERE "user_subscription__row"."id" = '5'
```

lib/DDGC/Schema/Result/User/Subscription.pm
```diff
@@ -0,0 +1,86 @@
...
+                "$activity.for_user" => { -ident => "$subscription.users_id" },
+            ],
+            -or => [
+                "$activity.for_role" => undef,
+                "$activity.for_role" => {
+                    -in => \"SELECT role FROM user_role WHERE users_id = $subscription.users_id",
+                }
+            ]
+        ]
+    };
```
jbarrett: Looking a bit more like it now:

```sql
SELECT "me"."id", "me"."created", "me"."category", "me"."action", "me"."meta1", "me"."meta2", "me"."meta3", "me"."description", "me"."format", "me"."for_user", "me"."for_role" 
  FROM "user_subscription" "user_subscription__row" 
  JOIN "activity_feed" "me" 
    ON "me"."category" = "user_subscription__row"."category" AND ( "user_subscription__row"."action" IS NULL OR "user_subscription__row"."action" = "me"."action" ) AND ( "user_subscription__row"."meta1" IS NULL OR "me"."meta1" LIKE "user_subscription__row"."meta1" ) AND ( "user_subscription__row"."meta2" IS NULL OR "me"."meta2" LIKE "user_subscription__row"."meta2" ) AND ( "user_subscription__row"."meta3" IS NULL OR "me"."meta3" LIKE "user_subscription__row"."meta3" ) AND ( "me"."for_user" IS NULL OR "me"."for_user" = "user_subscription__row"."users_id" ) AND ( "me"."for_role" IS NULL OR "me"."for_role" IN ( 
        SELECT role 
          FROM user_role 
        WHERE users_id = user_subscription__row.users_id
       ) ) 
WHERE "user_subscription__row"."id" = '5'
```

jbarrett:
@kablamo @nilnilnil I plan to merge this today (effectively rendering 'dev-community' in need of stabilisation) and start staging the changes, if there are no objections.

lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -395,6 +395,53 @@ many_to_many 'users', 'instant_answer_users', 'user';
 has_many 'instant_answer_topics', 'DDGC::DB::Result::InstantAnswer::Topics', 'instant_answer_id';
 many_to_many 'topics', 'instant_answer_topics', 'topic';
 
+after insert => sub {
+    my ( $self ) = @_;
+    my $schema = $self->result_source->schema;
+    $schema->resultset('ActivityFeed')->create(
+        $self->ddgc->config->subscriptions->created_ia_page( {
```
kablamo: Why not have the defaults here?  Probably because they are used in more than one place? If so, then how about adding a method to `DDGC::DB::Result::ActivityFeed` called .. `create_for_new_ia_page()` (there must be a better name but I can't think of one).
```
$schema->resultset('ActivityFeed')->create_for_new_ia_page({meta1 => ..., meta2 => ..., description => ...});
```

lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -395,6 +395,53 @@ many_to_many 'users', 'instant_answer_users', 'user';
...
+            meta1        => $self->id,
+            meta2        => join('', map { sprintf ':%s:', $_ }
+                $self->topics->columns([qw/ name /])->all),
```
kablamo:
```
:$topic::$topic::$topic:
```
huh.  what is logic behind this?  So you can grep for `:$topic:`?  Thats clever.  Its kind of a mouthful to type though.  Maybe you could make this a helper method on the `$self->topics` object.  related: The pod in `DDGC::Schema::Result::ActivityFeed` for this is different.

kablamo: Hmm I wonder if you could put a bunch of this logic into the `$schema->resultset('ActivityFeed')->create_for_new_ia_page()` method I was suggesting above.  Not sure.

jbarrett: Yeah, it's a delimeter so you can search for '%:$topic:%' and be sure you have matched a complete topic name.

:+1: Helper sounds like a plan, thanks for heads up on pod.

nilnilnil:
:+1:

lib/DDGC/Schema/ResultSet.pm
```diff
@@ -13,14 +13,21 @@ sub current_user {
...
 __PACKAGE__->load_components(qw/
     Helper::ResultSet::Me
-    Helper::ResultSet::Shortcut::HRI
-    Helper::ResultSet::Shortcut::Prefetch
-    Helper::ResultSet::Shortcut::GroupBy
-    Helper::ResultSet::Shortcut::OrderBy
-    Helper::ResultSet::Shortcut::Page
-    Helper::ResultSet::Shortcut::Rows
+    Helper::ResultSet::Shortcut
+    Helper::ResultSet::OneRow
```
kablamo: nice.  i did not know `first()` leaves a cursor open.  wonder how that affects postgres.  never had a problem with it though.  that i knew about.

jbarrett: Yeah, we use `->first` all here over and we also happen to leak lots of Cursor instances.

I prefer this approach for explicitly adding `LIMIT 1` to the query.

jbarrett: I also didn't know about it until this week - I think it's a relatively new addition.

PR #741, opened 2015-06-11
Styling fixes

jbarrett:
Small fixes to blog application styling, templates and login behaviour.


jbarrett:
@jagtalon Since we're not currently using SCEditor, can you confirm the css linking changes here are appropriate? Thanks!

https://github.com/duckduckgo/community-platform/commit/0998d40

jagtalon:
@jbarrett yes, that looks good! That same for-loop has been moved to https://github.com/duckduckgo/community-platform/blob/dev/src/scss/content/main.scss which just combines all of those files into one.

jbarrett:
@jagtalon Great, thanks a lot!

PR #746, opened 2015-06-11
Update bbcode help

jbarrett:
@zekiel Rather than faff about with tag cusomisation, I updated the help :)


zekiel:
I'm no mathematologist but when you subtract a negative, it's a positive.

PR #747, opened 2015-06-11
update IA page if pr changed

jdorweiler:
If the issue is a PR and there is already an existing IA page for it:
1. get all the usual fields that we're trying to auto create.
2a. If the existing IA page doesn't have anything for that field.  Update it.
2b. If the existing IA page already has data for that field, Skip it

For example, this will cover the cases where someone updates the PR later to do things like add api documentation, add a description if where wasn't already one, or adds a duck.co link.

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -87,9 +87,10 @@ sub getIssues{
                 my $data = \%entry;
                 return unless $data->{name};
 
-
+                # check to see if we have this IA already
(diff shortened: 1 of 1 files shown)


jdorweiler:
Requested by @moollaza 
cc: @MariagraziaAlastra @russellholt

MariagraziaAlastra:
Thanks @jdorweiler! Tested, LGTM :+1:

PR #752, opened 2015-06-12
Wear page updates.

jbarrett:


jbarrett:
Will repoen this against community branch

PR #753, opened 2015-06-15
Blog post authoring validation and feedback.

jbarrett:
- Validation fixes
- Retain errors for reporting in 'edit' template


jbarrett:
@jagtalon Just confirming (since you write the most posts I think) 'company blog' always ends up ticked, right?

PR #754, opened 2015-06-16
Redesign the static live page

jagtalon:
Currently playing around. Not ready for a review. Might be alot of changes.

[alot2](http://hyperboleandahalf.blogspot.com/2010/04/alot-is-better-than-you-at-everything.html)

**Changes so far:**

- The newest Litestrap version that's hosted in https://github.com/duckduckgo/litestrap has classes such as `.show--screen-l` or `.forty--screen-xs` available which I think will be useful in the redesign. Let's use that version. More details in the "Responsiveness" section of https://duckduckgo.com/styleguide.

- Added new classes in a file called `_sep.scss` for easily separating text with vertical bars.

screen shot 2015-06-16 at 9 18 10 am

- Include Litestrap's [button classes](https://github.com/duckduckgo/litestrap/blob/master/_buttons.scss) for nice buttons.

- There's now more space between the content and the headers.

- Commas are now added via CSS instead of being in HTML.

screen shot 2015-06-16 at 10 25 16 am

- Issues section is fixed. 

screen shot 2015-06-16 at 4 21 01 pm

**Guide:**

ia detail - interim


jagtalon:
Ok initial version is working. Kinda looks like the mockup now which I'm happy about. Now I think it's time to break it.

jagtalon:
Talked to @chrismorast @MariagraziaAlastra @russellholt, and here are their opinions:

- There could be too many issues. Limit the number of issues to, say, 5, and add a button that expands to show everything else. Should probably be sorted by importance, too.

- Screenshot should not include the search bar.

- The generate screenshot button should probably be overlayed on the image or at least the part where the image will go.

jagtalon:
Also I have to take care of how the content flows when the browser is resized.

jagtalon:
Some mockups for the screenshot button from @chrismorast :

screen shot 2015-06-16 at 2 11 32 pm

screen shot 2015-06-16 at 1 57 27 pm

MariagraziaAlastra:
@jagtalon I agree that for the first one we would probably be better off with an actual button, same style as "create an issue" button. I very much like the second mockup, though!

jagtalon:
This should be good for now. Will continue the rest of the issues in https://github.com/duckduckgo/community-platform/pull/754#issuecomment-112558253 in another PR.

MariagraziaAlastra:
:+1:

PR #761, opened 2015-06-18
IA Index - add ability to filter using URL params

MariagraziaAlastra:
@russellholt The URLs will be like "/ia?&repo=spice&template=info"; when the page is loaded the final part of the URL path gets splitted and for each field-value pair, if it's valid, a click event is triggered on the corresponding element.
It basically mimics the manual clicks on the filter options in the sidebar.

I just need to change the URL whenever a filter option is clicked, so we easily get a shareable link without the hassle of having to craft it ourselves, and then it's done.

### DIFF
--- a/lib/DDGC.pm
+++ b/lib/DDGC.pm
@@ -402,7 +402,12 @@ sub _build_xslate {
 				$source =~ s/$from/$to/g;
 				return $source;
 			},
-			urify => sub { lc(join('-',split(/\s+/,join(' ',@_)))) },
+			urify => sub {
+                my $value = shift;
+                $value = lc $value;
+                $value =~ s/[^a-zA-Z]+/-/g;
+                return $value;
+            },
 
 			floor => sub { floor($_[0]) },
 			ceil => sub { ceil($_[0]) },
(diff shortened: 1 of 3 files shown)


MariagraziaAlastra:
@russellholt [Screencast](http://gfycat.com/UnderstatedGregariousErmine)

russellholt:
:+1: 

- trailing `#` seems weird
- should revisit use of topic numbers vs. names in the future

MariagraziaAlastra:
@russellholt [Topic names as param values](http://gfycat.com/SilentHarmoniousAmericanriverotter)

The trailing "#" gets appended because the options are void links, aka anchor tags with "#" as href.

lib/DDGC.pm
```diff
@@ -402,7 +402,12 @@ sub _build_xslate {
 				$source =~ s/$from/$to/g;
 				return $source;
 			},
-			urify => sub { lc(join('-',split(/\s+/,join(' ',@_)))) },
+			urify => sub {
```
MariagraziaAlastra: Thanks @jbarrett, I would have never found this by myself!

russellholt:
very awesome.

PR #765, opened 2015-06-19
.gitignore: don't ignore built ddgc.css and ia.css.

jagtalon:
For @jbarrett


jbarrett:
:+1: Thanks @jagtalon 

I don't think we need ia.css

jbarrett:
@jagtalon Will this change make `grunt commit` add that css to the repo?

jagtalon:
@jbarrett If I'm not mistaken, `grunt commit` adds new files to the repo (the versioned files). But since this file is not versioned, and is already in the git repo, `grunt commit` shouldn't do anything to it. tl;dr it should be fine. :)

jbarrett:
To be clear, I'd like `grunt commit` to commit this as it's one less thing to do during release (a build is part of release, so there would be potential changes, right?)

Thanks!

jagtalon:
@jbarrett not sure if I understand, but it's already committed https://github.com/duckduckgo/community-platform/blob/jag/unignore/root/static/css/ddgc.css so there's no need for `grunt commit` to commit it?

jagtalon:
@jbarrett ready

jbarrett:
LGTM, thanks @jagtalon!

So simple in the end :)

PR #773, opened 2015-06-22
Overview page - exclude IA Pages with undefined dates

MariagraziaAlastra:
@russellholt Simple fix for excluding old IA Pages which don't have dates
screenshot-maria duckduckgo com 5001 2015-06-22 18-46-46


russellholt:
- should put launch (completed) dates by the live IAs
- no need to sort alphabetically the new ones alphabetically, though I assume those are all on the same day

MariagraziaAlastra:
@russellholt They're just sorted by date. The fact they're sorted alphabetically in this case is probably because those pages, as you said, were created on the same day.

PR #778, opened 2015-06-23
text changes to footer

zekiel:
//cc @jbarrett


templates/wear.tx
```diff
@@ -116,7 +118,7 @@
 	</div>
 </div>
 <: } :>
-
+to
```
jbarrett: I think this line is a typo?

zekiel: yep

PR #779, opened 2015-06-23
Do not pass ddg links through meta refresh redir

jbarrett:


bsstoner:
Pretty edge case, but should we worry about cases where `https://duckduckgo.com` might be in the querystring? i.e. `http://something.com?url=http://duckduckgo.com`

andrey-p:
The regex I put together on the spread page could be useful: `^https?:\/\/(\w+\.)?(duckduckgo\.com|duck\.co)`

bsstoner:
Also what about non-https links to `http://duckduckgo.com`? The regex might be a good idea.

You know better than me where this is used and how urgent it is. I'm just trying to spot potential holes. Feel free to merge if you're cool with it :)

jbarrett:
I'm going to go ahead with this, since it fixes an issue we have now and the caveats aren't current issues, going to open a new PR to properly deal with this since I woud like this to be robust.

PR #780, opened 2015-06-23
Perform redirect for hosts that aren't this and ddg

jbarrett:
@bsstoner @andrey-p Many thanks for you feedback in #779 

I think this should be closer to achieving what we want (filter out duckduckgo and $self, where $self is staging, dev or view - dynamic).

Could you also confirm I am right with my string escaping - I think I need to escape the backslashes themselves when building a new re from strings - this is what my REPL tells me anyway:

```
> /^https?:\/\/(\w+\.)?(duckduckgo\.com|duck\.co)/
/^https?:\/\/(\w+\.)?(duckduckgo\.com|duck\.co)/
> new RegExp("^https?:\\/\\/(\\w+\\.)?(duckduckgo\\.com|duck\\.co)")
/^https?:\/\/(\w+\.)?(duckduckgo\.com|duck\.co)/
```

Thanks!

### DIFF
--- a/src/js/ddgc/ddgc.js
+++ b/src/js/ddgc/ddgc.js
@@ -54,7 +54,7 @@ $(document).ready(function() {
 
 	$("a[href^='http']").mousedown(function() {
 		current_hostname = location.hostname;
-		current_hostname.replace(/\./g, '\\.');
+		current_hostname.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
 		hostname_re = new RegExp("^https?:\\/\\/(\\w+\\.)?(duckduckgo\\.com|" + current_hostname + ")");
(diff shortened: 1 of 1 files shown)


jbarrett:
Actually, I get the feeling I should be using something like what's described in the first answer here

http://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex

andrey-p:
Was just going to write that yeah, escaping those backslashes with extra backslashes is necessary - but using that function in the SO answer should work pretty well.

One unrelated thing I noticed - `location.hostname` will include the subdomain, and that should be covered by the `(\w+\.)` bit in the regex. (It checks for any number of alphanumeric characters followed by a dot.)

jbarrett:
Subdomain is useful for links to image. or if someone inserts www. - thanks!

PR #786, opened 2015-06-25
dont update any existing data from the db

jdorweiler:
When auto creating IA pages the ID was being changed whenever the name changed.  This caused the duplicated IA pages.  It also caused the problem that @zachthompson saw where he updated the name and it reverted back the next time ghIssues ran. 

To fix this I'm getting the IA page from the DB and only updating `undef` fields.  Any fields from the DB that are already set will remain unchanged. 

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -90,9 +90,6 @@ sub getIssues{
                 # check to see if we have this IA already
                 # First lookup by ID.  This can fail if an admin updates the ID on the IA page later
                 my $ia = $d->rs('InstantAnswer')->find($data->{name}, {result_class => 'DBIx::Class::ResultClass::HashRefInflator'}) || {};
-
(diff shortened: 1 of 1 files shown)


russellholt:
:+1:

MariagraziaAlastra:
@jdorweiler I got this error message while executing the script: <pre>
Use of uninitialized value $ENV{"DDGC_PROSODY_USERHOST"} in string ne at /home/ubuntu/community-platform/script/../lib/DDGC.pm line 31.
</pre>

MariagraziaAlastra:
Anyway it works - just want to know what that message is about and see if we can do something before merging. Thanks!

MariagraziaAlastra:
Merging, a fix for that error is already in dev-community (thanks @jbarrett).
Great job @jdorweiler! :+1:

PR #787, opened 2015-06-26
Allow numbers in blog title -> url conversion

jbarrett:


jbarrett:
Thanks to @b1ake for digging out a bunch of issues - PRs incoming!

PR #788, opened 2015-06-26
Update blog links in templates

jbarrett:
@zekiel This change also deprecates the OpenID setting - I think it has ~20 users.

It will continue to work for people who have configured it already (**if** indeed it actually works).


jbarrett:
Gah, I really wish you could change target branch on an open PR

PR #790, opened 2015-06-26
Deprecating 'Open links in new window' user setting.

jbarrett:


jbarrett:
This is in preparation for moving all markup rendering to `DDGC::Util::Markup`

PR #792, opened 2015-06-29
Fix generated HTML content tag closing

jbarrett:
`HTML::TreeBuilder::LibXML::Node::as_XML` doesn't handle empty elements correctly so if, say, some BBCode has empty nodes and generates `<b></b>`, `as_XML` will output `<b />`

`as_HTML` will output the correct opening / closing tags for empty nodes, but will also do the same for tags which don't nest, like `<br>`, `<hr>` and ``, i.e. you'll get `<br></br>`.

Since only `</br>` poses an actual problem in our output, I've filtered these out as a quick fix until other options are explored.

### DIFF
--- a/lib/DDGC/Util/Markup.pm
+++ b/lib/DDGC/Util/Markup.pm
@@ -202,7 +202,15 @@ sub html {
     my $guts = $tree->guts;
 
     if ($guts) {
-        return $guts->as_XML;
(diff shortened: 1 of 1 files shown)


jbarrett:
Wrong target branch. Howcome I never get this right?

PR #798, opened 2015-06-29
github.handlebars: Fend against missing tags.

jagtalon:
If there are no tags, position the "created" label somewhere else.

screen shot 2015-06-29 at 4 23 19 pm


MariagraziaAlastra:
:+1:

PR #801, opened 2015-06-30
Eric/community stats

kablamo:
This pr is for the Github community stats dashboard.  

# Back end

The existing code was written to populate the `github_*` tables, but that work wasn't completed.  This pr finishes some of that work and refactors some of the code.  I've also added a few tables and columns.

`github_*` tables are separate from the `instant_answer_*` tables.  The github tables represent the data from github while the instant_answer tables represent an instant answer.  

The github tables will be a convenient source of data for multiple projects including eventually the instant_answer tables.  For now its the data source for the github stats dashboard. In the future it will either be a data source for the instant_answer tables or the instant_answer table could link to the github tables directly via an id.

**Code**
 - `DDG::GitHub`
 - `DDGC::DB::Result::GitHub::*`
 - `DDGC::DB::ResultSet::GitHub::*`
 - `script/ddgc_import_github.pl`  <-- cron job to import db data

**How to setup the db**
 1. Update your db schema: 'sql/1.044/github_tables.sql`
 2. To import data from the github api create a token [here] (https://github.com/settings/tokens)
 3. Add the token to your environment: `export DDGC_GITHUB_TOKEN=_your_token_here_`
 4. Load data into tables from the github api: `script/ddgc_import_github.pl`

# Business logic

The logic which calculates the stats. 

**Code**
 - `DDGC::Stats::GitHub`
 - `DDGC::Stats::GitHub::Utils`

# Front end

**controller:** `lib/DDGC/Web/Controller/InstantAnswer.pm`
**template:** `/templates/instantanswer/dashboard.tx`
**path:** `/ia/dashboard`

This page is intended to be used internally and requires an admin account to access.  Later on these kinds of stats will be exposed on the ia pages.  But exactly how that will work is undecided.

This is the first version of this dashboard.  The plan is to add graphs to right hand side in the next version.  That will make it easier to comprehend the data and compare for example last week with other previous weeks.

Layout inspired by https://status.github.com/


kablamo:
@MariagraziaAlastra I'm guessing you are interested in this pr.  It might be easeir if I walk you through it.

MariagraziaAlastra:
Hey @kablamo!
Sure, I'd also like to involve @russellholt, @jdorweiler and @jbarrett if they want/can.

In the meantime could you write an high level overview of the problem you're trying to solve, what this code does and how it interacts (if it does) with the existing codebase, what's missing/next steps etc?

I think it would be helpful both as a reference and also in order to have everyone on the same page, before getting into the actual code review. Thanks!

kablamo:
Sure!  This is still a wip, but the more feedback the better.  I've updated the description.

MariagraziaAlastra:
Nice writeup, thanks! I just noticed the PR is to the dev branch, though - right now for IA Pages we're making PRs to dev-ia-pages, while John works on the complat architecture on the dev-community branch, and we use dev as integration branch.

@jbarrett do you think we can merge this PR to dev, or does it need be moved to dev-ia-pages? 

Anyway @kablamo I'll do some testing later, thank you again :)

lib/DDGC/GitHub.pm
```diff
@@ -8,25 +8,35 @@ use DateTime::Format::ISO8601;
...
+
+    # something to do with users giving us access to their github accounts via oath
+    $github->validate_session_code($code, $user);
+
+    # find or update github_user table for this user
+    $github->find_or_update_user($login);
+
+=cut
+
+has ddgc       => (isa => 'DDGC', is => 'ro', weak_ref => 1, required => 1);
```
kablamo: @jbarrett I think you mentioned this to me as being a problem before.  Passing `$ddgc` around makes everything very tightly coupled.  How should I handle it?  

I think I should change this module ui to work like this 
```
my $github = DDGC::GitHub->new(
    db => $db,           # required; a DDGC::DB obj
);
```
and instantiate my own copy of DDGC::Config inside this module.  Ideally DDGC::Config would be singleton I guess?

lib/DDGC/DB/Result/GitHub/Comment.pm
```diff
@@ -0,0 +1,42 @@
...
+column github_user_id      => {data_type => 'bigint', is_nullable => 0};
+column number              => {data_type => 'bigint', is_nullable => 0};
+column body                => {data_type => 'text',   is_nullable => 0};
+column created_at          => {data_type => 'timestamp without time zone', is_nullable => 0};
+column updated_at          => {data_type => 'timestamp without time zone', is_nullable => 0};
+column gh_data             => {
+    data_type        => 'text',
+    is_nullable      => 0,
+    serializer_class => 'AnyJSON',
+    default_value    => '{}',
```
MariagraziaAlastra: @kablamo does this prevent errors when calling from_json and to_json in case the string is empty? Just guessing

Also, what about the AnyJSON serializer class? What does it do? I think we use the JSON serializer class elsewhere...

kablamo: Yeah, but since the field is not nullable, it should never get used.  Probably I don't need that line.

The AnyJSON serializer appears to be a custom serializer which is defined [here](https://github.com/duckduckgo/community-platform/blob/master/lib/DBIx/Class/InflateColumn/Serializer/AnyJSON.pm).  I think I should be using the standard JSON serializer.

MariagraziaAlastra:
@kablamo Could you add a sql file with the schema changes? We usually create a folder with the next release version number inside sql/ and then put there all the scripts and sql which need to run in order to update the DB: https://github.com/duckduckgo/community-platform/tree/master/sql

lib/DDGC/DB/Result/GitHub/Comment.pm
MariagraziaAlastra: Alright, thanks for explaining :)

kablamo:
@MariagraziaAlastra Sure.  I just added the schema changes [here](https://github.com/duckduckgo/community-platform/pull/801/files?diff=split#diff-7b36c6415551416745f63fccb199cce5).

MariagraziaAlastra:
@kablamo when I try to update the DB on my instance using the sql file I get this:
<pre>BEGIN
psql:sql/1.044/github_tables.sql:17: NOTICE:  CREATE TABLE will create implicit sequence "github_comment_id_seq" for serial column "github_comment.id"
psql:sql/1.044/github_tables.sql:17: NOTICE:  CREATE TABLE / PRIMARY KEY will create implicit index "github_comment_pkey" for table "github_comment"
psql:sql/1.044/github_tables.sql:17: NOTICE:  CREATE TABLE / UNIQUE will create implicit index "github_comment_github_id" for table "github_comment"
CREATE TABLE
psql:sql/1.044/github_tables.sql:20: ERROR:  there is no unique constraint matching given keys for referenced table "github_issue"
psql:sql/1.044/github_tables.sql:23: ERROR:  current transaction is aborted, commands ignored until end of transaction block
</pre>

kablamo:
@MariagraziaAlastra sorry about that!  I blindly took the output from `script/ddgc_schema_diff.pl` and assumed that would work.  I've changed the order of statements and removed some things I didn't mean to have in there.  Try it again!

MariagraziaAlastra:
Now it works, thanks @kablamo!
When I get to the ia/dashboard page I get this error, though:
<pre>Caught exception in DDGC::Web::Controller::InstantAnswer->dashboard "Illegal division by zero at /home/ubuntu/community-platform/script/../lib/DDGC/Stats/GitHub/PullRequests.pm line 41."</pre>

kablamo:
@MariagraziaAlastra thats because the tables are empty.  Sorry I should have added this stuff to the description.  You need to run 
```
script/ddgc_import_github.pl
```
to populate the tables.

MariagraziaAlastra:
I think I still have something missing, or I don't know what, because when running that script I get the following:
<pre>Undef did not pass type constraint "Str" (in $args->{"access_token"}) at (eval 1521) line 97
    "Str" is a subtype of "Value"
    "Value" is a subtype of "Defined"
    Undef did not pass type constraint "Defined" (in $args->{"access_token"})
    "Defined" is defined as: (defined($_))
</pre>

kablamo:
@MariagraziaAlastra I forgot about that.  You are missing the DDGC_GITHUB_TOKEN environment variable.  This environment variable already exists on the production server.  But its better security if you create and use your own rather than using that one.
1. To import data from the github api create a token [here] (https://github.com/settings/tokens)
 2. Add the token to your environment: `export DDGC_GITHUB_TOKEN=_your_token_here_`

MariagraziaAlastra:
@kablamo Did all that and when I run the script now I get:
<pre>encountered object '0', but neither allow_blessed, convert_blessed nor allow_tags settings are enabled (or TO_JSON/FREEZE method missing) at /opt/perlbrew/perls/perl-5.16.3/lib/site_perl/5.16.3/JSON.pm line 154.
DBIx::Class::Storage::TxnScopeGuard::DESTROY(): A DBIx::Class::Storage::TxnScopeGuard went out of scope without explicit commit or error. Rolling back. at /opt/perlbrew/perls/perl-5.16.3/lib/site_perl/5.16.3/JSON.pm line 154
</pre>

lib/DDGC/GitHub.pm
jbarrett: Yeah, DDGC::Config is usually accessed from the DDGC instance. So, having DDGC instances passed into application components doesn't have to be a problem, since the application components live as long as the application so circular refs don't really matter. We do end up with leaks elsewhere because of DBIC abuse.

In this case, I would just pass a `ddgc` ref as that happens to be easiest - you then get ->config and ->rs.

My problem with this generally is the separation of concerns gets really fuzzy, for example a template can send email by calling methods of `$c.d.postman` - this has led to a tendency where chunks of functionality exist across all layers rather than, say, adhering to fat model. It has also earned us some pretty bad leaks.

For now, having a weak ref to `ddgc` is how DDGC is designed to work so go with the flow :smiley:

kablamo:
@MariagraziaAlastra the problem was you had an older version of Net::GitHub than I did.  I updated the cpanfile to require v0.77.  I also installed the new version for you in your `~/community-platform/local` lib directory.

MariagraziaAlastra:
Hey @kablamo, sorry, I still have the same error I posted in my last comment when I run the script :(

MariagraziaAlastra:
No wait, it's slightly different: I don't get the DBIx part, I get only the first line about allow_blessed etc

MariagraziaAlastra:
Update: it's working now, thanks to @kablamo; I had to add ./local/lib/perl5 to my PERL5LIB before running the script, using
<pre>PERL5LIB=./local/lib/perl5:./lib:$PERL5LIB perl -d:Confess script/ddgc_import_github.pl</pre>

Note: `perl -d:Confess` forces a stack trace in case there are errors; it enables the module `Devel::Confess`, so it's necessary to set the PERL5LIB env variable, in order to use it.

kablamo:
@MariagraziaAlastra  @jbarrett I'm done adding features here and would like to release what I've got.  If you could look it over when you get some time that would be great.  Note that I've updated the schema since you last looked at it so the sql schema script needs to be run again as well as `ddgc_import_github.pl`.  You probably need to truncate all the `github_*` tables before running the import script.  I've also moved the dashboard page to `/admin/stats/github/last_week`.

Known issues:
1. Travis is failing and I need to look into that. 
2. my schema script now has a bad version #. Not sure where the version # comes from.  It seems to mysteriously increment.

I'm sure there is other stuff.  Feedback welcome!

lib/DDGC/DB/Result/GitHub/ReviewComment.pm
```diff
@@ -0,0 +1,43 @@
...
+use Moose;
+extends 'DDGC::DB::Base::Result';
+use DBIx::Class::Candy;
+use namespace::autoclean;
+
+table 'github_review_comment';
+
+primary_column id         => {data_type => 'bigint', is_auto_increment => 1};
+unique_column  github_id  => {data_type => 'bigint', is_nullable => 0};
+column github_repo_id     => {data_type => 'bigint', is_nullable => 0};
```
jbarrett: Nice - I should really actually **read** that Candy doc :)

templates/admin/stats/github.tx
```diff
@@ -0,0 +1,108 @@
...
+
+    /* colors */
+    h2 {
+        color : #aaa;
+    }
+
+    /* state */
+    .loader {
+        visibility : hidden;
+    }
```
jbarrett: Is it worth putting this into external css?

MariagraziaAlastra: @kablamo You could use the "hide" class instead. Anyway yeah please export the style to an scss file inside src/scss/ia/, and import it in [src/scss/ia/main.scss](https://github.com/duckduckgo/community-platform/blob/master/src/scss/ia/main.scss) - also, instead of the hex colours, you can use [variables](https://github.com/duckduckgo/duckduckgo-colors/blob/master/_colors.scss) :)

lib/DDGC/Web/Controller/Admin/Stats.pm
```diff
@@ -246,5 +247,45 @@ sub coupons :Chained('base') :Args(0) {
...
+    $c->stash->{stats} = \@stats;
+    $c->stash->{tabs}  = {
+        last_24_hours    => "",
+        last_week        => "",
+        week_before      => "",
+        week_before_that => "",
+        last_month       => "",
+        last_90_days     => "",
+    };
+    $c->stash->{tabs}->{$since} = "selected";
```
MariagraziaAlastra: @kablamo This page should be only visible to admins, right? I don't see any check for that if that's the case. Maybe redirect to /ia if user isn't logged in or is not an admin?

jbarrett:
Noting new for Pinto:

KAZUHO/Cache-LRU-0.04.tar.gz
FAYLAND/Net-GitHub-0.77.tar.gz

@kablamo Been looking at /ia/dashboard an embarrassingly long amount of time thinking "No way he *forgot* to commit the controller and template".

Anyway, functionally LGTM :+1:

Assuming @zekiel has been brought up to speed?

lib/DDGC/Web/Controller/Admin/Stats.pm
jbarrett: Anything chained from base in this controller is in turn chained from the /admin/base which should perform this authorisation check.

Starting to dislike this chained dispatcher business (well, I say starting...)

https://en.wikipedia.org/wiki/COMEFROM :smile:

templates/admin/stats/github.tx
kablamo: Oh yeah.  Forgot.  I was going to ask how I should do this.  Thanks for pointing it out.  Oh cool its scss.

lib/DDGC/Web/Controller/Admin/Stats.pm
MariagraziaAlastra: Alright, cool, thanks @jbarrett!

kablamo:
@jbarrett 
- Yup, I showed it to @zekiel yesterday and he is copied on the deploy task. 
- Sorry forgot to update the pr description with the right path
- Oh I was going to ask about pkgs too: Is the cpanfile not being used at all?  If not, why is it in there?  Good point about adding them to Pinto.  I will do that.

jbarrett:
Not sure why tests are failing - they pass in dev, so...

cpanfile is used on deploy, but the pinto mirror is specified - we don't use carton or such.

kablamo:
@jbarrett what uses the cpanfile if you don't use carton?

jbarrett:
@kablamo https://github.com/duckduckgo/community-platform/blob/0dc490e/script/ddgc_release.sh#L49 - cpanm

MariagraziaAlastra:
@kablamo I removed all the data and I got the DB updated correctly now (I hope) but maybe not, because when I run script/ddgc_import_github.pl I get:

<pre>Updating github_user table
Undef did not pass type constraint "Str" (in $args->{"access_token"}) at (eval 1563) line 97
    "Str" is a subtype of "Value"
    "Value" is a subtype of "Defined"
    Undef did not pass type constraint "Defined" (in $args->{"access_token"})
    "Defined" is defined as: (defined($_))
</pre>

Any idea of why this happens?

jbarrett:
@MariagraziaAlastra You may need to update your dependencies:

`dzil listdeps --missing | grep -v abstract | cpanm` should do it.

MariagraziaAlastra:
@jbarrett Thanks, but I still get that error :(

jbarrett:
OH, D'oh! Should have **read** your error message @MariagraziaAlastra - apologies!

So, you need to have a GitHub API ket in your environment - it describes how to do this in the task description up top.

MariagraziaAlastra:
@jbarrett Yeah, I exported it again - I think I did that about ten times, not sure why I need to do this again everytime - and now it looks like it's working!

MariagraziaAlastra:
@kablamo I'm finally able to look at the stats page, and it's nice! Just a minor wording suggestion:
screenshot-maria duckduckgo com 5001 2015-07-16 16-03-42

Instead of "Week before" and "Week before that", what do you think of having "Two weeks ago" and "Three weeks ago"?
As I said, just a minor suggestion. Great job, LGTM! :+1:

kablamo:
@MariagraziaAlastra thats a good idea.  

Probably you already are aware of this but maybe the reason you have to keep exporting the variable is because you opened a new terminal?  All env variables get reset when you open a new terminal.  If you always want that variable to be around, you have to add it to your .bashrc.

MariagraziaAlastra:
@kablamo Yeah I just did that, thanks! :)

kablamo:
Before changes to .travis.yml tests took 19 minutes and 24 minutes.
After changes to .travis.yml tests took 15 minutes and 13 minutes.
Not amazing, but a little better.  

We could remove another 10 minutes of that time by caching dependencies.  But it wasn't super easy to figure out so I'm not gonna do it now.
http://docs.travis-ci.com/user/caching/

PR #803, opened 2015-07-01
InstantAnswer.pm - set test machine to null when changing dev milestone to 'live'

MariagraziaAlastra:
@russellholt @jagtalon as requested


jagtalon:
@MariagraziaAlastra can we have an SQL script to retroactively do this for the IAs that were already set to live? Prob not important though, but I could imagine _something_ (or _someone_) to make use of that information, and I think keeping our DB up-to-date would be for our benefit.

PR #805, opened 2015-07-01
IAIndex - fix query URL param

MariagraziaAlastra:
@russellholt fixes the regression on the query param.
Leaving this open, I want to remove the first "&" as well


russellholt:
LGTM

MariagraziaAlastra:
Done with it, merging

PR #784, opened 2015-06-25
Object instance leak detection

jbarrett:
@nilnilnil @malbin 

This is intended to help track down object instance leaks in a coarse fashion. The idea is you pick a route and do multiple runs against it with different numbers of iterations (GET requests) and compare the numbers:

```
$ script/ddgc_leak_report.pl 1 '/help' 2>&1 | grep -v ^DEP
Tracked objects by class:
        DBD::Pg::DefaultValue                    1
        DBI::var                                 5
        DBIx::Class::Cursor::Cached              1
        DBIx::Class::Storage::DBI::Cursor        1
        Hash::Merge                              2
$ script/ddgc_leak_report.pl 100 '/help' 2>&1 | grep -v ^DEP
Tracked objects by class:
        DBD::Pg::DefaultValue                    1
        DBI::var                                 5
        DBIx::Class::Cursor::Cached              100
        DBIx::Class::Storage::DBI::Cursor        100
        Hash::Merge                              2
```

Excuse the grep - the new applications have erroneous deprecation warnings. Anyway, you can see we are accumulating instances of DBIC Cursor modules in the help controller.

PR #806, opened 2015-07-02
Deprecate XSlate results() calls

jbarrett:
[`results()`](https://github.com/duckduckgo/community-platform/blob/19cdc66/lib/DDGC.pm#L373-L381)

- This appears to leak DBIC stuff (see #784)
- Let's sort in the RDBMS instead, it's very good at it.

[`Devel::Leak::Object`](https://metacpan.org/pod/Devel::Leak::Object) profiling shows none of the DBIC instances we saw before hanging around after this change, so let's do the same for everything practical:

```
$ script/ddgc_leak_report.pl 1 /help 2>&1 | grep -v ^DEP
Tracked objects by class:
        DBD::Pg::DefaultValue                    1
        DBI::var                                 5
        Hash::Merge                              2
$ script/ddgc_leak_report.pl 100 /help 2>&1 | grep -v ^DEP
Tracked objects by class:
        DBD::Pg::DefaultValue                    1
        DBI::var                                 5
        Hash::Merge                              2
```

Even if this fails to reduce our footprint, it should at least reduce load a smidgen.

// @nilnilnil @malbin @zekiel

### DIFF
--- a/lib/DDGC.pm
+++ b/lib/DDGC.pm
@@ -369,17 +369,6 @@ sub _build_xslate {
 			# Mark text as raw HTML
 			r => sub { mark_raw(@_) },
 
-			# trick function for DBIx::Class::ResultSet
-			results => sub {
-				my ( $rs, $sorting, $order ) = @_;
-				my @results = $rs->all;
-				$sorting
-					? ( ($order // '') eq 'asc' )
-						? [ sort { $a->$sorting <=> $b->$sorting } @results ]
-						: [ sort { $b->$sorting <=> $a->$sorting } @results ]
-					: [ @results ];
-			},
-
 			# general functions avoiding xslates problems
 			call => sub {
 				my $thing = shift;
--- a/lib/DDGC/DB/Base/ResultSet.pm
+++ b/lib/DDGC/DB/Base/ResultSet.pm
@@ -43,4 +43,8 @@ sub prefetch_context_config {
   return \%prefetch;
 }
 
+sub all_ref {
+  [ $_[0]->all ];
+}
+
 1;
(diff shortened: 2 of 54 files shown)


jbarrett:
20 down, 27 to go.

jbarrett:
So there are some improvements here, going by the object instance profiler, but we're not out of the woods yet.

**Before**

```
$ script/ddgc_leak_report.pl 1 /forum/general 2>&1 | grep -v ^DEP
Tracked objects by class:
        DBD::Pg::DefaultValue                    1
        DBI::var                                 5
        DBIx::Class::Cursor::Cached              1
        DBIx::Class::Storage::DBI::Cursor        2
        Hash::Merge                              2
$ script/ddgc_leak_report.pl 5 /forum/general 2>&1 | grep -v ^DEP
Tracked objects by class:
        DBD::Pg::DefaultValue                    1
        DBI::var                                 5
        DBIx::Class::Cursor::Cached              5
        DBIx::Class::Storage::DBI::Cursor        10
        Hash::Merge                              2
```

**After**

```
$ script/ddgc_leak_report.pl 1 /forum/general 2>&1 | grep -v ^DEP
Tracked objects by class:
        DBD::Pg::DefaultValue                    1
        DBI::var                                 5
        DBIx::Class::Storage::DBI::Cursor        1
        Hash::Merge                              2
$ script/ddgc_leak_report.pl 5 /forum/general 2>&1 | grep -v ^DEP                                                                                                                       
Tracked objects by class:
        DBD::Pg::DefaultValue                    1
        DBI::var                                 5
        DBIx::Class::Storage::DBI::Cursor        5
        Hash::Merge                              2
```

I suspect (but currently have no evidence) that the forum leak is in [`DDGC::Web::Table`](https://github.com/duckduckgo/community-platform/blob/c051264/lib/DDGC/Web/Table.pm) as I think that's where its resultset gets built - see the pager stuff.

jbarrett:
Currently on https://view.dukgo.com/

PR #809, opened 2015-07-02
IA Pages - make sure the ID (and meta_id) doesn't have weird chars in it before saving

MariagraziaAlastra:
@russellholt @jagtalon the check happens when creating a new IA Page and also when editing the meta_id both on the dev pages and the live pages; any non-alphanumeric char gets replaced by an underscore, unless they're at the beginning or end of the string: in that case they get cut off entirely.

The ID can have numbers in it, but they can't be at the beginning of the string, and they need to be mixed with alphabetic (a-z) chars.

If the string only contains weird chars and/or numbers, it gets emptied, so the caller won't save it.

Also, now inside the dialog for creating a new IA Page the ID input field turns red if the entered string is invalid and thus the page is not created.

screenshot-maria duckduckgo com 5001 2015-07-02 17-13-51

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -878,14 +878,14 @@ sub create_ia :Chained('base') :PathPart('create') :Args() {
 sub format_id {
     my( $id ) = @_;
 
-    # id must be unique, lowercase and without weird chars
-    $id =~ s/[^a-zA-Z0-9]+/_/g;
-    $id =~ s/^[^a-zA-Z]//;
+    # id must be lowercase and without weird chars
+    $id = lc $id;
+    $id =~ s/[^a-z0-9]+/_/g;
+    $id =~ s/^[^a-zA-Z]+//;
     $id =~ s/[^a-zA-Z0-9]$//;
 
     # make the id string empty if it only contains non-alphabetic chars
     $id =~ s/^[^a-zA-Z]+$//;
-    $id = lc $id;
 
     return $id;
 }
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -838,6 +838,7 @@ sub create_ia :Chained('base') :PathPart('create') :Args() {
     my $ia = $c->d->rs('InstantAnswer')->find({id => lc($c->req->params->{id})}) || $c->d->rs('InstantAnswer')->find({meta_id => lc($c->req->params->{id})});
     my $is_admin;
     my $result = '';
+    my $id = '';
 
     if ($c->user && (!$ia)) {
        $is_admin = $c->user->admin;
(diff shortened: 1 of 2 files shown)


lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -848,34 +846,50 @@ sub create_ia :Chained('base') :PathPart('create') :Args() {
...
     $c->stash->{not_last_url} = 1;
     return $c->forward($c->view('JSON'));
 }
 
+sub format_id {
+    my( $id ) = @_;
+
+    # id must be unique, lowercase and without weird chars
+    $id =~ s/[^a-zA-Z0-9]+/_/g;
+    $id =~ s/^[^a-zA-Z]//;
```
jagtalon: This makes "123rain" into "23rain"

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -848,34 +846,50 @@ sub create_ia :Chained('base') :PathPart('create') :Args() {
...
+    $id =~ s/[^a-zA-Z0-9]$//;
+
+    # make the id string empty if it only contains non-alphabetic chars
+    $id =~ s/^[^a-zA-Z]+$//;
+    $id = lc $id;
```
jagtalon: Maybe do `lc` first so that you don't have to do `[a-zA-Z]`

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -848,34 +846,50 @@ sub create_ia :Chained('base') :PathPart('create') :Args() {
...
     };
 
     $c->stash->{not_last_url} = 1;
     return $c->forward($c->view('JSON'));
 }
 
+sub format_id {
+    my( $id ) = @_;
+
+    # id must be unique, lowercase and without weird chars
```
jagtalon: How do the regexes below check if an ID is unique?

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -848,34 +846,50 @@ sub create_ia :Chained('base') :PathPart('create') :Args() {
...
+    $id =~ s/[^a-zA-Z0-9]+/_/g;
+    $id =~ s/^[^a-zA-Z]//;
+    $id =~ s/[^a-zA-Z0-9]$//;
```
jagtalon: Wouldn't this only remove one non-alphanumeric character at the end?

jagtalon:
Why isn't "rain_bow" being accepted?

lib/DDGC/Web/Controller/InstantAnswer.pm
MariagraziaAlastra: Oh haha that's checked in the caller function, I cut and pasted the comment from there, sorry :)

lib/DDGC/Web/Controller/InstantAnswer.pm
MariagraziaAlastra: Yeah I just want to totally remove them at the end (because the first line replaces all of them with underscores, so if there's something at the end of the string we still need to remove that last underscore)

MariagraziaAlastra:
I'll investigate, thanks for the feedback @jagtalon!

MariagraziaAlastra:
@jagtalon Try to break it again :)
Anyway rain_bow is accepted on my instance, if you have the same problem please let me know (unless you were trying to save an already existent ID and that's why it wasn't accepted?)

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -848,34 +846,50 @@ sub create_ia :Chained('base') :PathPart('create') :Args() {
...
 }
 
+sub format_id {
+    my( $id ) = @_;
+
+    # id must be lowercase and without weird chars
+    $id = lc $id;
+    $id =~ s/[^a-z0-9]+/_/g;
+    $id =~ s/^[^a-zA-Z]+//;
+    $id =~ s/[^a-zA-Z0-9]$//;
```
jagtalon: Should this be `s/[^a-zA-Z0-9]+$//`?

MariagraziaAlastra: I'm not doing that because in the first regex all the consecutive non-alphanumeric chars are replaced by just one underscore, so for example "test)&%" would become "test_" and that line makes sure we remove that final underscore.

jagtalon: @MariagraziaAlastra Alright. What about `s/_$//` then?

MariagraziaAlastra: :+1:

MariagraziaAlastra: @jagtalon done :cactus:

jagtalon:
👌

MariagraziaAlastra:
:rocket:

PR #811, opened 2015-07-02
Deprecate DDGC::Web::Table

jbarrett:
Following on from #806, Web::Table's job is rendering pagers for rendered resultsets. Problems:

- It appears to leak.
- It's more complex than just doing counting / paging by hand.

So rather than fix it, I'm just going to delete it. It's used in ~~four~~ three views:

- Forum
- Ideas
- ~~Moderation and Reports~~
- Userpage comments

(Moderations invoked it but didn't actually **use** it)


jbarrett:
Something else happening here.

PR #817, opened 2015-07-09
Array-ref sigil typo

jbarrett:
@MariagraziaAlastra Can you confirm this still does what's expected? Thanks.


MariagraziaAlastra:
Thanks @jbarrett, LGTM :+1:

PR #826, opened 2015-07-15
Upgrade Net::GitHub to v0.77

kablamo:
Arg.  @jdorweiler  I figured out why I could not run your ghIssues.pl script on my box -- because I upgraded to Net::GitHub v0.77 which handles utf8 strings differently so you no longer need to decode stuff.

But I need v0.77 because it has methods for new parts of the github api.  So I think we need to release an updated version of your script at the same time I upgrade Net::GitHub.

This pr has the changes.  

cc @MariagraziaAlastra @jbarrett

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -11,7 +11,6 @@ use Data::Dumper;
 use Try::Tiny;
 use Net::GitHub;
 use Time::Local;
-use Encode qw(decode_utf8);
 my $d = DDGC->new;
 
 # JSON response from GH API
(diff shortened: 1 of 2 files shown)


MariagraziaAlastra:
@jdorweiler is out this week!

script/ghIssues.pl
```diff
@@ -74,8 +73,8 @@ sub getIssues{
 				repo => $repo || '',
 				issue_id => $issue->{'number'} || '',
                 author => $issue->{user}->{login} || '', 
-				title => decode_utf8($issue->{'title'}) || '',
-				body => decode_utf8($issue->{'body'}) || '',
+				title => $issue->{'title'} || '',
+				body => $issue->{'body'} || '',
```
MariagraziaAlastra: @kablamo So what this means is this new New::Github version automatically gets values decoded from utf8?

kablamo: yup

MariagraziaAlastra: Nice, will test this later!

MariagraziaAlastra:
@kablamo Looks good :+1:

PR #829, opened 2015-07-17
rm styling for h1, h2, and a becuase it affects other pages

kablamo:
cc @MariagraziaAlastra @jbarrett


MariagraziaAlastra:
Thanks @kablamo, LGTM! :+1:

PR #836, opened 2015-07-22
Ideas: Push the votes back to the right.

jagtalon:
Old:

screen shot 2015-07-22 at 2 12 33 pm

New:

screen shot 2015-07-22 at 2 12 41 pm

### DIFF
--- a/root/static/css/content.css
+++ b/root/static/css/content.css
@@ -1,3 +1,4 @@
+@charset "UTF-8";
 /* DDG Compass Vars */
 /* for compass */
 /*
@@ -4862,6 +4863,7 @@ Buttons/Tags
 
 /* line 669, ../../src/scss/content/_forms-n-buttons.scss */
 .vote {
+  float: right;
   display: inline-block;
   border: 1px solid #c6c6c6;
   -moz-border-radius: 3px;
@@ -4872,7 +4874,7 @@ Buttons/Tags
   position: relative;
 }
 
-/* line 679, ../../src/scss/content/_forms-n-buttons.scss */
+/* line 680, ../../src/scss/content/_forms-n-buttons.scss */
 .vote__button {
   float: none;
   height: 28px;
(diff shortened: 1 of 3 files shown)


jagtalon:
screen shot 2015-07-22 at 2 14 01 pm

screen shot 2015-07-22 at 2 14 07 pm

jbarrett:
I think I'd rather figure out where these rules went.

I need some help digging into history since I don't understand css, scss or the build system.

I've attempted to bisect from 0dc490e (which is good, release 1.058) - attempting to rebuild the css for each commit is not a valid approach since some revisions do not build, so I just went through with incomplete CSS looking at "vote on left" vs "vote on right", ignoring other stuff. That came up with 2ed0659 - would appreciate any help you can provide reviewing this @jagtalon @MariagraziaAlastra 

Something to do with those removed media rules? They appear to be in use on forum and idea pages.

MariagraziaAlastra:
@jbarrett Yeah it's definitely those rules, especially [this one](https://github.com/duckduckgo/community-platform/commit/2ed06595d527d81e8c9861c9debadc9287398702#diff-64a612c667686eae6bf77ba7505abd89L393).
I guess then it's just a matter of putting those rules back in place.

jbarrett:
I'd like to understand how they went missing too... but yeah, just put them back in place to see...

Yup

jbarrett:
linking #837

MariagraziaAlastra:
Right, @jagtalon did you remove them on purpose or was that by accident?
I should have tested that PR more extensively, too.

jbarrett:
I wonder if we can hijack @andrey-p's work to automate this kind of evaluation? I think we'd need actial CI first, right?

jagtalon:
@MariagraziaAlastra @jbarrett https://github.com/duckduckgo/community-platform/pull/838

PR #839, opened 2015-07-24
Only redirect HTTP(s) URLs.

jbarrett:


lib/DDGC/Web/Controller/Root.pm
```diff
@@ -113,7 +113,12 @@ sub redirect_duckco :Chained('base') :PathPart('topic') :Args(1) {
 sub redir :Chained('base') :PathPart('redir') :Args(0) {
 	my ( $self, $c ) = @_;
 	$c->stash->{not_last_url} = 1;
-	$c->session->{r_url} = $c->req->param('u');
+	my $u = URI->new( $c->req->param('u') );
+	if ( $u->scheme !~ /^http/i ) {
```
nilnilnil: What about using if

nilnilnil: Damn you Mobile.

Uri->secure

jbarrett: We also use this for plain HTTP links, so it's not suitable in this instance unfortunately.

nilnilnil:
:+1:

PR #840, opened 2015-07-27
Eric/community stats

kablamo:
This is some one off code I used to generate some stats for Zac.  If Zac's experiment works, then I'll be doing more work on this.  If not, then this pr was perhaps a waste of time.


jbarrett:
Cool if I leave this hang until tomorrow? Sorry for delay...

kablamo:
No rush.

lib/DDGC/Stats/Comleader.pm
```diff
@@ -0,0 +1,141 @@
...
+            context      => [qw/DDGC::DB::Result::Idea DDGC::DB::Result::Thread/],
+        }, { 
+            select => [{ count => 'me.id', -as => 'comment_count' }, 'users_id'],
+        })
+        ->group_by('users_id')
+        ->order_by({ -desc => 'comment_count' });
+
+    while (my $comment = $rs->next) {
+        next unless $comment->user->public;
+        next if     $comment->user->ghosted;
```
jbarrett: I wouldn't worry too much about the user themselves being ghosted if they happen to have a lot of approved comments.

lib/DDGC/Stats/Comleader.pm
```diff
@@ -0,0 +1,141 @@
...
+            select   => [ { count => 'me.id', -as => 'translation_count' }, 'translator_users_id' ],
+            group_by => 'translator_users_id',
+        })
+        ->order_by({ -desc => 'translation_count' });
+
+    while (my $token_language = $rs->next) {
+        my $user = $token_language->translator_user || next;
+
+        next unless $user->public;
+        next if $user->ghosted;
```
jbarrett: Same deal here - if their translations are going live, we maybe don't care to much about the user flag.

`$user->ignore` is maybe a much stronger signal to replace this with - this happens to presistent spammers.

jbarrett:
Besides notes above, :+1:

lib/DDGC/Stats/Comleader.pm
kablamo: what do you mean by approved comments? do you mean comments which are not ghosted?  seems like we would not want to make a comleader out of someone who has been hellbanned?

jbarrett: So everyone is ghosted by default. Removing this flag requires explicit action on the part of the mods so we can't be confident it has been done in every case.

PR #841, opened 2015-07-28
"Claim" an IA Idea

jbarrett:


jbarrett:
@zekiel @russellholt 

So the most basic UI for this would be a "Develop this idea!" or such button which makes you the claimant. We can make this deletable by admins or community leaders if it remains claimed for too long and someone wants to jump in?

jbarrett:
Sounds like this is already in the pipeline elsewhere, will reopen this if required.

russellholt:
Let's talk about it wednesday.

lib/DDGC/Web/Controller/Ideas.pm
```diff
@@ -186,6 +186,26 @@ sub idea : Chained('idea_id') PathPart('') Args(1) {
...
+		} );
+	}
+	elsif ( $c->stash->{idea}->claimed_by &&
+	     ( $c->stash->{idea}->claimed_by == $c->user->id ||
+	       $c->user->is('admin') ) ) {
+		$c->stash->{idea}->update( {
+			claimed_by => undef,
+		} );
+	}
+	$c->response->redirect( $c->chained_uri(@{ $c->stash->{idea}->u }) );
```
MariagraziaAlastra: I guess the question here is: after we create an IA Page for this idea (and I think the claim button shouldn't exist until that happens) and a user claims it, should we update the developer field for that IA Page as well? I think we should.
Even though, if the user ends up doing nothing it will be a little bit of a pain to remove him/her from the developer array...

jbarrett:
So we're missing somewhere to indicate *who* has claimed but:

Unclaimed

This button on the right is visible to everybody.

Claimed

This button is visible to the... "claimant" and admins. Should probably make it comleaders too.

Thinking the name who staked the claim should be in the title on the post page and an extra piece of info somewhere in the index. Open to suggestions.

lib/DDGC/Web/Controller/Ideas.pm
jbarrett: We could consider "claiming" as an expression of interest. Perhaps an interstitial page should say "Talk to us about starting development of this IA" with a link back to the IA idea?

Not sure how this fits with the work flow, but it acts as a gateway for people who are interested in developing to get into the process perhaps?

MariagraziaAlastra:
@jbarrett what about the idea thread status? Maybe it should automatically change to "in development" and it shouldn't be possible to set it back to "needs a developer" unless it's unclaimed?

jbarrett:
Or "Needs a developer" sets it back to unclaimed?

MariagraziaAlastra:
Yeah that too

lib/DDGC/Web/Controller/Ideas.pm
russellholt: I think the claim button has to notify an admin.
The admin, when viewing the idea, then has a "create IA" button on the idea.

MariagraziaAlastra: I agree on notifying an admin, but I think the "create IA" button on the idea should be there before. What if we don't like this idea? Creating the page means we approve it and so it can be taken by a developer.

jbarrett: Hmmm, perhaps it should only be available for certain statuses? (Needs a developer, Quack and hack...)

russellholt: That's why only an admin creates the IA, once the developer asks to claim it.

jbarrett:
I think this is a reasonable minimal approach we can expand on later? Assigning to @MariagraziaAlastra - thanks!

MariagraziaAlastra:
@jbarrett  LGTM :+1:

PR #843, opened 2015-07-29
Quack and Hack IA status

jbarrett:
- Adds 'Quack and Hack' status.
- Adds ability to link to status by name rather than numeric ID (e.g. /ideas/status/quack-and-hack, /ideas/status/live)
- Can be reverted to "Needs a Developer" status after the event with a single db op


jbarrett:
@zekiel @tagawa

PR #845, opened 2015-08-04
Only present claim button to users with <= 5 claims

jbarrett:


jbarrett:
Other stuff coming here, perhaps. Stay tuned.

PR #847, opened 2015-08-04
capitalize Instant Answer

moollaza:
We capitalize it everywhere in the docs and we should follow suite here too. It also helps to make the term a little more pronounced on the page.

to @jbarrett 
/cc @zekiel


jbarrett:
:+1:

PR #848, opened 2015-08-05
Only present claim button to users with <= 5 claims

hubub:


jbarrett:
@hubub Is there something I can help you with?

PR #861, opened 2015-08-12
_dev.scss: Prevent example query editing from capitalizing

jagtalon:


MariagraziaAlastra:
Thanks @jagtalon, LGTM :+1:

PR #869, opened 2015-08-18
Merge idea manager / forum manager data migration clauses

jbarrett:
Not doing this will lead to pkey conflicts since there is so much overlap in the sets


jbarrett:
This has been tested on view.

PR #871, opened 2015-08-19
Suppresses Moose warnings with new DBIC.

jbarrett:
cc @kablamo - I think you spotted this in travis output


kablamo:
Ah yes.  Well I spotted it on my dev box.  I'm not 100% sure I understand the problem.  But thats the fix I did as well.

lgtm :+1:

PR #874, opened 2015-08-19
Don't update the users table.  Its not needed atm.

kablamo:
# Overview

Updating the github_user table causes problems when the user has deleted their account or renamed it.  Need to restructure the code to handle that.  But atm this feature isn't needed.  So I will just comment it out.  This has the added benefit of decreasing the # of requests to GitHub api by like 80%.  

# Testing

 * [x] Ran script/ddgc_import_github.pl manually on my dev box.

### DIFF
--- a/lib/DDGC/GitHub.pm
+++ b/lib/DDGC/GitHub.pm
@@ -95,8 +95,8 @@ sub one_second {
 sub update_database {
     my ($self) = @_;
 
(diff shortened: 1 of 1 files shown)


kablamo:
cc @MariagraziaAlastra

jbarrett:
:+1:

jbarrett:
This is staged on view.

PR #875, opened 2015-08-19
missing tab names

jdorweiler:
Fix missing tab names


jdorweiler:
copied over and ran it manually

PR #881, opened 2015-08-20
Subscribe a user to IA Page updates when IA is claimed

jbarrett:
cc @zekiel @russellholt @MariagraziaAlastra @nilnilnil 

I discussed this with @MariagraziaAlastra already, would be interested in other thoughts.

First thought of mine, doing this without being explicit about it is potentially problematic (especially without a nice way to unsubscribe or remove the subscription).

### DIFF
--- a/lib/DDGC/Schema/Result/User.pm
+++ b/lib/DDGC/Schema/Result/User.pm
@@ -4,6 +4,8 @@ package DDGC::Schema::Result::User;
 
 use Moo;
 extends 'DDGC::Schema::Result';
+with 'DDGC::Schema::Role::Result::User::Subscription';
+
 use DBIx::Class::Candy;
 use Scalar::Util qw/ looks_like_number /;
 use namespace::autoclean;
--- a/lib/DDGC/DB/Base/ResultSet.pm
+++ b/lib/DDGC/DB/Base/ResultSet.pm
@@ -10,6 +10,7 @@ extends qw(
   DBIx::Class::Helper::ResultSet::Shortcut
   DBIx::Class::Helper::ResultSet::CorrelateRelationship
   DBIx::Class::Helper::ResultSet::SetOperations
+  DBIx::Class::Helper::ResultSet::OneRow
 );
 
 sub ddgc { shift->result_source->schema->ddgc }
--- a/lib/DDGC/DB/Result/Idea.pm
+++ b/lib/DDGC/DB/Result/Idea.pm
@@ -221,6 +221,13 @@ sub toggle_claim {
 		return 1;
 	}
 	elsif ( $self->claimed_by == $user->id || $user->is('forum_manager') ) {
+		my $claimant = ( $self->claimed_by == $user->id )
+		    ? $user
+		    : $self->schema->resultset('User')->find( $self->claimed_by );
+		my $ia = $self->instant_answer;
+		if ( $claimant && $ia ) {
+			$claimant->unsubscribe_from_instant_answer( $ia->id );
+		}
 		$self->update( {
 			claimed_by => undef,
 		} );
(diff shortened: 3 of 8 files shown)


lib/DDGC/Schema/Result/User.pm
```diff
@@ -4,6 +4,8 @@ package DDGC::Schema::Result::User;
 
 use Moo;
 extends 'DDGC::Schema::Result';
+with 'DDGC::Schema::Role::Result::User::Subscription';
+
```
jbarrett: @kablamo I'd be interested in your thoughts on this - pulling related stuff for a given result class out to a role.

I like it since, while I like "fat model", I dislike these multi-thousand line classes. I'm just not clear on the... idiomaticness of having highly specialised (i.e. non-reusable) roles.

kablamo: I also dislike long classes.  

What I did once was pull out all the functions for these Result classes into Roles.  I decided that code and schema definition were 2 separate things and belonged in two separate files.  I kind of liked that.  So there was one role for a corresponding Result class and the role was definitely not reusable.  

I'm a strong believer that doing something just for readability is enough of a justification to do it.  I make lots of small new functions that are only used once.

kablamo: I guess another way to make these classes shorter is to put the column definition all on one line.  I find that a lot more readable when looking at the schema definition.  But it usually violates the 80 char limit rule.  Which is not cool.  My justification is that readability is more important and rules are made to be broken.  Definitely a personal preference thing.  :)

jbarrett: I genuinely prefer reading the multi-line column defs - I think I get bored if my eyes have to stray to far to the right :wink: ...though if there are dozens of them, that is a sign that something else may be amiss. Still, rather that than these damned JSON / YAML serialised columns we have. I did spot your single line column defs in the Github work, it does *look* very tidy.

I can see these Result classes become column defs and a bunch of `with` specifications, but there comes a point when too many files is as rough to deal with as long files. I hope I can keep my scatterbrain on an appropriately coarse granularity.

Thanks for input!

kablamo: > I can see these Result classes become column defs and a bunch of with specifications, but
> there comes a point when too many files is as rough to deal with as long files.

To clarify, I just had one `with` statement, not a bunch of `with` statements.  I put all the subs in that one role.  So the dir structure might look like this:
```
.
├── Result
│   ├── Account.pm        # schema definition
│   ├── User.pm             # schema definition
│   └── Role
│        ├── Account.pm   # methods for Result::Account
│        └── User.pm        # methods for Result::User
├── ResultSet
│   ├── Account.pm
│   └── User.pm
...
```
That way I always knew to look for the schema definitions in the `Result/` dir and to look for the methods for Result classes in `Result/Role` dir.  Maybe thats still too many files for you.  Whatever floats your boat!

jbarrett: Makes sense, but I think having a two-thousand line role is the same problem as a large result class so plan to not explicitly eschew the idea of composing result classes from multiple roles.

zekiel:
not exactly sure what's going on here but general concept of subbing to IA page notifs after claiming is good. 

Wondering if this PR has anything related to the current_Follow button or any other Follow ability (if not the claimant)

russellholt:
if the user unclaims...  leave the notification in place?

jbarrett:
Actually, unclaim could be an ideal unsubscribe step.

jbarrett:
@zekiel I guess we can see what feedback we get, if any. This is unrelated to anything currently called "Follow"  - no subscription to the idea is created, just the page.

jbarrett:
@russellholt @zekiel In the absence of a frontend for this, I made the "Unclaim" action also unsub the user from updates on the related IA page.

zekiel:
:+1:

PR #883, opened 2015-08-20
Advanced Activity Rendering

jbarrett:
So the emails we've recieved so far have been pretty uninformative. "IA Page Created" is OK, but "IA Page Updated" is less useful - what exactly has been updated?

This role allows **optional** advanced rendering for an activity feed entry. If description isn't going to cut it, you can expand on it here with a sub based on your category name and action.

...I am beginning to think that an application that wants to qualify as `$schema->app` should provide an xslate instance since instantiating one for each result (whether needed or not) is pretty crusty and probably slow.

...though not as slow as everything else :wink:

### DIFF
--- a/lib/DDGC/Schema/Role/Result/ActivityFeed/AdvancedDescription.pm
+++ b/lib/DDGC/Schema/Role/Result/ActivityFeed/AdvancedDescription.pm
@@ -1,0 +1,40 @@
+package DDGC::Schema::Role::Result::ActivityFeed::AdvancedDescription;
+use strict;
+use warnings;
+
+use Moo::Role;
+
+has xslate => (
+    is => 'ro',
+    lazy => 1,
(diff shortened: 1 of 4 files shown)


lib/DDGC/Schema/Role/Result/ActivityFeed/AdvancedDescription.pm
```diff
@@ -0,0 +1,40 @@
...
+        path => 'views',
+    );
+}
+
+sub describe {
+    my ( $self ) = @_;
+    my $sub = sprintf 'describe_%s',
+                join( '_', ( $self->category, $self->action ) );
+
+    return $self->$sub if $self->can( $sub );
```
jbarrett: Holy metacrap, Batman!

I think this is the only way to get the flexibility I'd like for this.

PR #884, opened 2015-08-20
Gruntfile.js: Don't let deleteBuildFiles error out.

jagtalon:
When first running Grunt, `build/` might not exist so deleteBuildFiles throws an error. This workaround makes sure that there's a build/ directory in the first place before doing anything.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -220,7 +220,7 @@ module.exports = function(grunt) {
         exec: {
             revert: "./script/revert_pkg_version.pl",
             revert_release: "./script/revert_pkg_version.pl release",
-            deleteBuildFiles: "rm -r build",
+            deleteBuildFiles: "mkdir -p build && rm -r build",
             bower: "bower install"
         },
 


jagtalon:
It prevents this from happening:

```
[11:38 AM jag@jag community-platform {master}]$ grunt
Running "exec:bower" (exec) task

Running "exec:deleteBuildFiles" (exec) task
>> rm: cannot remove `build'
>> : No such file or directory
>> Exited with code: 1.
Warning: Task "exec:deleteBuildFiles" failed. Use --force to continue.

Aborted due to warnings.
```

jagtalon:
For @jdorweiler or @jbarrett CC @MariagraziaAlastra

Gruntfile.js
```diff
@@ -220,7 +220,7 @@ module.exports = function(grunt) {
         exec: {
             revert: "./script/revert_pkg_version.pl",
             revert_release: "./script/revert_pkg_version.pl release",
-            deleteBuildFiles: "rm -r build",
+            deleteBuildFiles: "mkdir -p build && rm -r build",
```
jbarrett: I guess this works, but it kind of seems perverse to create / delete, how about:

```bash
[ -d build ] && rm -r build
```

I think the intent is more  clear this way?

jagtalon: @jbarrett that is great thank you

jagtalon: This doesn't work because `[ -d build ] && rm -r build` will return exit 1 if `build/` doesn't exist.

jbarrett: O well, do the thing that works so :)

PR #888, opened 2015-08-20
fix messed up perl modules from ghIssues

jdorweiler:
I'm pretty sure this is what was causing the messed up perl modules like `DDG::Sports/MLB/Games`.

I don't have the complat set up so this needs some testing.

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -117,6 +117,8 @@ sub getIssues{
                     $tmp_repo =~ s/s$//g;
 
                     if(my ($name) = $file->{filename} =~ /lib\/DDG\/$tmp_repo\/(.+)\.pm/i ){
+                        my @parts = split('/', $name);
+                        $name = join('::', @parts);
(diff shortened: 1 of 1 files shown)


jbarrett:
@jdorweiler What are you missing on your setup? Not that much involved in getting up and running.

script/ghIssues.pl
```diff
@@ -117,6 +117,8 @@ sub getIssues{
                     $tmp_repo =~ s/s$//g;
 
                     if(my ($name) = $file->{filename} =~ /lib\/DDG\/$tmp_repo\/(.+)\.pm/i ){
+                        my @parts = split('/', $name);
+                        $name = join('::', @parts);
```
jbarrett: Not that it matters, but FWIW you can chain split and join:

```perl
$ join '::', split '/', 'Sports/MLB/Games'
Sports::MLB::Games
```

jdorweiler:
Thanks @jbarrett

PR #890, opened 2015-08-21
Don't generate activity from automated activity redux

jbarrett:


jbarrett:
@jdorweiler @MariagraziaAlastra I'm a little confused about the updates I'm getting from view - every hour there are about 40 IAs updated twice but the code here should stop activity updates happening.

Is there anything else updating IA pages on a schedule like this?

jbarrett:
Oh wait, I'm an idiot - these updates would have been generated in the hour before I deployed.

Never mind!

PR #894, opened 2015-08-25
Fix C3 Method Resolution errors.

jbarrett:
For some reason the DDGC::DB base ResultSet class was written to extend the DBIC components it used rather than use the `load_components` method.

Using `load_components` we no longer get conflicts around `Helper::ResultSet::Shortcut::OrderBy` and `Helper::ResultSet::Shortcut::OrderByMagic`.

Further work may include restricting Shortcut helpers to the set we use but we shouldn't **have to** do this.

### DIFF
--- a/lib/DDGC/DB/Base/ResultSet.pm
+++ b/lib/DDGC/DB/Base/ResultSet.pm
@@ -4,14 +4,15 @@ package DDGC::DB::Base::ResultSet;
 use Moose;
 use namespace::autoclean;
 
-extends qw(
-  DBIx::Class::ResultSet
-  DBIx::Class::Helper::ResultSet::Me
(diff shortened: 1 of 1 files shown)


jbarrett:
@malbin @jkv This is our root cause.

kablamo:
I don't see those warnings any more.  Nice catch.  I think this might fix those warnings we saw and 'fixed' previously.   The ones about `Not inlining 'new'` or whatever.

http://stackoverflow.com/questions/22483145/how-to-setup-a-dbixclass-schema-with-moose-the-definite-guide

:+1:  lgtm

jbarrett:
Thanks @kablamo! Merging for release today.

PR #895, opened 2015-08-25
Deprecate has_flag - user roles no longer stored in serialised column

jbarrett:
cc @kablamo - Thanks!


kablamo:
nice.  thanks for that.

PR #896, opened 2015-08-25
Experimenting with composing multiple schemas from common roles / functionality

jbarrett:


jbarrett:
Going to try explain what I intend here until the code reveals itself and so I don't forget my intent.

Every unit of functionality in the project has some dependency on others via dependency injection or shared object instances. These means no component is reusable in other applications. It would be great, for example, if our Schema was a discrete, reusable unit. It depends on an instance of DDGC, which itself is a collection of other object instances (for everything from sending email to rendering templates to interfacing with XMPP to caching).

Some of this is obviously useful. For example, having an instance of the User result reflecting the currently logged in user allows us to query data directly based on their access level and more. Having the ability to render markdown or bbcode allows us to return prepared content straight from, e.g. Thread result class instances.

On the other hand, giving templates the ability to send email is more exposure than we strictly need - separation of concerns is weak.

A way to balance what's useful with what is perhaps too much access between components while also making components as reusable as possible is challenging. Let's take an example, [DDGC::Markup](https://github.com/duckduckgo/community-platform/blob/c3d225a/lib/DDGC/Markup.pm) - this mainly renders BBCode to HTML. This has a ref to a full DDGC instance mainly for access to two pieces of information, the image proxy URL and whether or not a user wants links to have a new window target. The latter having been removed from the settings page, all we need here is a single configuration value which could be set on instantiation. This module doesn't need to send email so having a DDGC instance is more exposure than is strictly required. We have far much more responsibility than strictly needed.

The DBIC Schema also has this exposure. A DDGC instance is used to access config, current_user, BBCode parsing but also everything else. This means any new application wishing to reuse the schema must either have a DDGC instance or make a lot of functionality available.

Since some of this functionality is genuinely useful, I propose a sort of "contract" to control the scale of the dependencies being passed around (just the raw number of instance references we need to keep track of) to a minimum. Separation of concerns.

What I mean by this is, we say "If you wish to use this schema as-is, you will need to provide some basic elements of functionality". The schema class itself could wrap these elements (see how `ddgc_config` is found in different applications), but if your $app instance dependency can't do it you probably can't use it... with a possible fall-back position of "We have decomposed the Result(Set) class definitions into a series of Roles. If you wish to compose your own schemas from this without the required functions, best of luck!"

The benefit here is we're able to, while factoring out what's not strictly necessary, re-marry the two Schemas in the repo (one of which was created because the other could not be re-used). Ultimately we may require different top-level schema classes for app interfacing but if the contract is reasonable and intact we should be able to reuse the Result(Set) definitions wholly.

Some top of the head maybe useful stuff in this contract:

- `uri_for` - host and application aware URL building
- `current_user` - an instance of the User Result for who is logged in
- Various rendering utils for markup, bbcode...
- Application configuration(?)

It's a pretty small set. It doesn't really need all of DDGC.

A potential side effect of not having these deep and wide dependency trees is we potentially reduce our leak and other performance issues. "Potentially" doing a lot of work here.

jbarrett:
"Consolidate schemas" PR incoming soonly

PR #900, opened 2015-08-26
claimed ideas without an IA page

jdorweiler:
This will find all the idea posts that are claimed but don't have an IA page.  Then create the IA page and subscribe the user to notifications for it.

### DIFF
--- a/script/check_claimed_ideas.pl
+++ b/script/check_claimed_ideas.pl
@@ -14,7 +14,7 @@ BEGIN {
 MAIN: {
     my $claimed_without_page = $d->rs('Idea')
         ->search({
-            claimed_by          => !undef,
+            claimed_by          => {'!=' => undef},
             instant_answer_id   => undef,
         });
 
--- a/script/check_claimed_ideas.pl
+++ b/script/check_claimed_ideas.pl
@@ -19,7 +19,7 @@ MAIN: {
         });
 
     my @time = localtime(time);
-    my $date = "$time[4]/$time[3]/".($time[5]+1900);
+    my $date = $d->db->format_datetime( DateTime->now );
 
     while (my $idea = $claimed_without_page->next){
 
@@ -44,6 +44,7 @@ MAIN: {
         $ia = $d->rs('InstantAnswer')->update_or_create({%ia_page});
         $idea->update({instant_answer_id => $ia_page{id}});
         $idea->user->subscribe_to_instant_answer( $ia->id );
+        $ia->add_to_users($idea->user);
 
         print "Created IA page: $ia_page{id}\n";
     }
--- a/script/check_claimed_ideas.pl
+++ b/script/check_claimed_ideas.pl
@@ -3,12 +3,7 @@
 # if not then make one
 use FindBin;
 use lib $FindBin::Dir . "/../lib";
-use JSON;
 use DDGC;
-use HTTP::Tiny;
-use Data::Dumper;
-use Try::Tiny;
-use Net::GitHub;
 use Time::Local;
 my $d = DDGC->new;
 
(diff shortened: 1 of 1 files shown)


jdorweiler:
Tested locally and on view.

script/check_claimed_ideas.pl
```diff
@@ -0,0 +1,50 @@
...
+
+MAIN: {
+    my $claimed_without_page = $d->rs('Idea')
+        ->search({
+            claimed_by          => {'!=' => undef},
+            instant_answer_id   => undef,
+        });
+
+    my @time = localtime(time);
+    my $date = "$time[4]/$time[3]/".($time[5]+1900);
```
russellholt: why 1900

script/check_claimed_ideas.pl
```diff
@@ -0,0 +1,50 @@
...
+            id              => $ia->{id} || $idea->id,
+            meta_id         => $ia->{meta_id} || $idea->id,
+            dev_milestone   => $ia->{dev_milestone} || 'planning',
+            name            => $ia->{name} || $idea->title,
+            description     => $ia->{description} ||$idea->content,
+            created_date    => $ia->{created_date} || $date,
+            forum_link      => $ia->{forum_link} || $idea->id,
+        );
+
+        $ia = $d->rs('InstantAnswer')->update_or_create({%ia_page});
```
russellholt: also what about `$ia->add_to_users($idea->user);`

script/check_claimed_ideas.pl
jdorweiler: The year `$time[5]` from localtime is the number of years since 1900.  
http://perldoc.perl.org/functions/localtime.html

russellholt: haha.. right

jbarrett: Generally we don't build date strings like that for SQL, since there is potential for ambiguity in day / month ordering and string formatting. There is a DateTime instance "parser" which should take care of these quirks for you:

```perl
my $date = $d->db->format_datetime( DateTime->now );
```

This will also pass a time to the RDBMS but it will truncate it.

script/check_claimed_ideas.pl
jdorweiler: added

script/check_claimed_ideas.pl
```diff
@@ -0,0 +1,51 @@
...
+    my @time = localtime(time);
+    my $date = $d->db->format_datetime( DateTime->now );
+
+    while (my $idea = $claimed_without_page->next){
+
+        my $ia = $d->rs('InstantAnswer')
+            ->find(
+                $idea->id,
+                {
+                    result_class =>  'DBIx::Class::ResultClass::HashRefInflator'
```
jbarrett: FYI, I'm pretty sure we have the DBIC helper configured to do HashRefInflator (via rs) like:

```perl
$d->rs('InstantAnswer')->search({ id => $idea->id })->hri;
```

Won't work with find, since that returns a Result instance.

script/check_claimed_ideas.pl
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env perl
+# check all claimed ideas and make sure they have an IA page.
+# if not then make one
+use FindBin;
+use lib $FindBin::Dir . "/../lib";
+use DDGC;
+use Time::Local;
```
jbarrett: I don't think this script will compile until you `use DateTime;`

jdorweiler:
@jbarrett thanks, should be good now

jbarrett:
:+1:

russellholt:
talked about the dates with @jdorweiler , looks good

:cactus:

PR #901, opened 2015-08-28
Fix idea status updating from ComLeaders

jbarrett:


jbarrett:
Hope to quash this "same code in two places" madness with work in #896 (and beyond)

PR #902, opened 2015-08-28
Last activity, comments, and commits

jdorweiler:
Added last activity date.
Also added last commit and last comment.  Here's what the data looks like.  Anything else I should get?

```
last_comment = {
    user => users name,
    date => date of comment,
    text => text of comment,
    id => id to link back 
};

last_commit = {
     diff => link to diff,
     user => user name,
     date => date,
     message => commit message,
 };

```

### DIFF
--- a/lib/DDGC/DB/Result/InstantAnswer.pm
+++ b/lib/DDGC/DB/Result/InstantAnswer.pm
@@ -385,6 +385,24 @@ column forum_link => {
     is_nullable => 1
 };
 
+column last_commit => {
+    data_type => 'text',
+    is_nullable => 1,
+    is_json => 1,
+};
+
+column last_comment => {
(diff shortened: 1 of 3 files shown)


jdorweiler:
@MariagraziaAlastra

MariagraziaAlastra:
I think that's all we need, thanks @jdorweiler!

russellholt:
LGTM :+1:

russellholt:
If this is good , tested - can we merge ?

MariagraziaAlastra:
It's not tested, I'm having issues with running the script on my instance (but it doesn't depend on this particular change)

MariagraziaAlastra:
Ok tested now, and it works! Thanks @jdorweiler! :+1:

PR #903, opened 2015-08-28
WIP: Add a search bar in the dev pipeline.

jagtalon:
The search field goes through each name of the IA in the dev pipeline and hides `<li>` elements that don't match.

For @MariagraziaAlastra

### DIFF
--- a/src/templates/dev_pipeline.handlebars
+++ b/src/templates/dev_pipeline.handlebars
@@ -1,4 +1,3 @@
-<input type="text" class="search-thing">
 <div>
 <div class="dev_pipeline-column left" id="pipeline-planning">
     <h2 class="dev_pipeline-column__title">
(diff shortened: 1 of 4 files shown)


src/templates/dev_pipeline.handlebars
```diff
@@ -1,3 +1,5 @@
+<input type="text" class="search-thing">
```
russellholt: I'll let @MariagraziaAlastra address this but you might want to call `search-thing` something more specific :palm_tree:

MariagraziaAlastra:
Tested and it works, nice job @jagtalon! Now let's make this a little bit prettier, more like the search box in the index. Also I'd like to see the search input placed on the same line of the pager button on top, but on the opposite side, I think it would save space and look cleaner as well. What do you think?

jagtalon:
@MariagraziaAlastra I think that would look good

jagtalon:
@MariagraziaAlastra made the changes!

MariagraziaAlastra:
Thanks @jagtalon! LGTM :+1:

jagtalon:
Thanks!

PR #905, opened 2015-08-29
Match id or meta_id from IA Page URI in ghIssues

jbarrett:

Thanks!


MariagraziaAlastra:
Looks good @jbarrett! Thanks! :+1:

PR #906, opened 2015-08-29
Reduce max claims to two

jbarrett:
Cc: @zekiel @tagawa 

This change is live, just submitting for permanent inclusion.


preemeijer:
oh no... ... ok :+1:

PR #907, opened 2015-08-29
Fix IA PR / issue linking

jbarrett:
Use retrieved IA id, not result name to update since result name can be meta_id.


jdorweiler:
:+1:  thanks for fixing @jbarrett

MariagraziaAlastra:
Thanks @jbarrett!

PR #908, opened 2015-08-29
last update fields isn't json

jdorweiler:


MariagraziaAlastra:
Thanks @jdorweiler :+1:

PR #909, opened 2015-08-29
ghIssues.pl - avoid rolling back updates when the IA doesn't exist

MariagraziaAlastra:
Thanks @jbarrett!
//cc @jdorweiler


jdorweiler:
thanks again for fixing ghissues :cactus:

PR #911, opened 2015-08-31
Ideas.pm - fix claimed page (duck.co/ideas/claimed)

MariagraziaAlastra:
@tagawa Just needed to get ideas which are claimed && have an IA Page, instead of claimed ideas with an undef IA Page :)

Now it works:
screenshot-maria duckduckgo com 5001 2015-08-31 16-33-19


tagawa:
Looks good -- thanks @MariagraziaAlastra

PR #920, opened 2015-09-03
Dev Pipeline - show elapsed time since last activity

MariagraziaAlastra:
Requires moment.js (install it using bower: `bower install`)

@russellholt 
screenshot-maria duckduckgo com 5001 2015-09-03 19-03-26

### DIFF
--- a/src/js/ia/Helpers.js
+++ b/src/js/ia/Helpers.js
@@ -4,7 +4,9 @@
     Handlebars.registerHelper("timeago", function(date) {
         if (date) {
             date = date.replace("T", " ").replace("Z", " ");
-            return moment(date, "YYYY-MM-DD HH:mm:ss Z").fromNow();   
+            date = moment(date, "YYYY-MM-DD HH:mm:ss Z").fromNow();
+            date = date.replace(/\sago/, "").replace(/a\s/, "1 ");
+            return date;
         }
     });
 
--- a/src/js/ia/Helpers.js
+++ b/src/js/ia/Helpers.js
@@ -1,5 +1,12 @@
 (function(env) {
     // Handlebars helpers for IA Pages
+    
+    Handlebars.registerHelper("timeago", function(date) {
+        if (date) {
+            date = date.replace("T", " ").replace("Z", " ");
+            return moment(date, "YYYY-MM-DD HH:mm:ss Z").fromNow();   
+        }
+    });
 
     /**
      * @function plural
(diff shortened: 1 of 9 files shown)


src/js/ia/IADevPipeline.js
```diff
@@ -16,6 +16,9 @@
                 var iadp;
                 iadp = Handlebars.templates.dev_pipeline(data.dev_milestones);
                 $("#dev_pipeline").html(iadp);
+
+                // 100% width
+                $(".site-main > .content-wrap").first().removeClass("content-wrap").addClass("wrap-pipeline");
```
russellholt: looks dangerous.

At least, in the future someone will be wondering why the pipeline behaves so differently while attempting to implement the redesign.

Is there a way we can note this fact so when the time comes we won't forget?

src/js/ia/Helpers.js
```diff
@@ -1,5 +1,14 @@
 (function(env) {
     // Handlebars helpers for IA Pages
+    
+    Handlebars.registerHelper("timeago", function(date) {
```
russellholt: comment about the expected input format. the `replace("T", " ")` stuff.

src/js/ia/IADevPipeline.js
MariagraziaAlastra: That's what we do in the single page as well, otherwise we couldn't have that blue band with 100% width.

I thought we agreed we need more space on the pipeline, at least for now that we only use the grid view.

How/where do you want to note this fact?

src/js/ia/Helpers.js
MariagraziaAlastra: Done

src/js/ia/IADevPipeline.js
russellholt: Yes, we agreed. Maybe a good place to note it is as a comment in the scss/css where `.content-wrap` is first defined. Like: 
```
// special note: .content-wrap is removed by pages x and y. see <files links>.
```

russellholt:
LGTM.

russellholt:
sorting?

MariagraziaAlastra:
Separate PR. Let me add that comment and the I'll merge

src/scss/ia/main.scss
```diff
@@ -19,3 +19,4 @@
 @import "_dev";
 @import "_live";
 @import "_overview";
+@import "_pipeline";
```
jagtalon: @MariagraziaAlastra I think you're missing a file?

MariagraziaAlastra: DARGH

MariagraziaAlastra: Added in another PR and merged, thanks @jagtalon!

PR #921, opened 2015-09-03
Dev Pipeline - sort by most recent activity

MariagraziaAlastra:
@russellholt 
screenshot-maria duckduckgo com 5001 2015-09-03 21-29-58


russellholt:
I think we really want to do the sorting in javascript, so that it can be reordered by a few metrics.
date of last activity (which is equivalent to this one - idle time), date of creation, and producer are three of the obvious ones.

For the prototyping purposes of *this week*, we can go with this. This is an example of blurring the lines between the front end and the back end, which is something we do not want. We will need to revisit this next week, and move the sorting to javascript on the front end.

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -138,7 +138,12 @@ sub dev_pipeline_json :Chained('dev_pipeline_base') :PathPart('json') :Args(0) {
 
     my @ias;
     my $key;
-    @ias = $rs->search({'dev_milestone' => { '=' => ['planning', 'development', 'testing', 'complete']}});
+    @ias = $rs->search(
```
russellholt: the future intentions for this code should be noted in a comment, why it's here, etc.

russellholt:
I haven't seen generated CSS going into the same PR like this here. Usually that goes into master, does it not?

russellholt:
@jagtalon can you test this please?

MariagraziaAlastra:
@russellholt I've chosen this solution because that's the easiest and the fastest to implement, for our prototyping scope of this week.

Of course that should be done on the front-end instead.

The generated css should be ignored by git automatically, I don't know what the deal is with it, I'll just remove that file.

lib/DDGC/Web/Controller/InstantAnswer.pm
MariagraziaAlastra: Alright

russellholt:
:+1:

PR #925, opened 2015-09-04
assign producer

jdorweiler:
Assign the producer to whoever is set as the github assignee.  Also one other small improvement I think you will like :smile: 
@MariagraziaAlastra

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -61,11 +61,9 @@ sub getIssues{
             # get the IA name from the link in the first comment
 			# Update this later for whatever format we decide on
 			my $name_from_link = '';
-
-            if($issue->{'body'} =~ /(http(s)?:\/\/(duck\.co|duckduckgo.com))?\/ia\/(view)?\/(\w+)/im){
-				$name_from_link = $5;
-			}
-			# remove special chars from title and body
+            ($name_from_link) = $issue->{'body'} =~ /https?:\/\/duck\.co\/ia\/view\/(\w+)/i;
+			
+            # remove special chars from title and body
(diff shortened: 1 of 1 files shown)


MariagraziaAlastra:
Awww, the progressbar!

script/ghIssues.pl
```diff
@@ -96,6 +103,7 @@ sub getIssues{
                 last_update => $issue->{updated_at},
                 last_commit => $last_commit,
                 last_comment => $last_comment,
+                producer => $issue->{assignee}->{login} || undef,
```
MariagraziaAlastra: @jdorweiler Apparently not a lot of PRs have an assignee, I'm getting a lot of undef producer values :(
Maybe fall back to @moollaza instead of undef? Or randomly choose between him and @jagtalon :smile: 

Just throwing ideas around!

jdorweiler: Well that would only matter for the IA pages.  If we really want I could randomly assign someone and then push that update to the PR on github too.

MariagraziaAlastra: That would be great, thanks!

jdorweiler:
Need to link github id to duck.co ia so moving it to WIP for now.

jdorweiler:
Ready to go again.  I'm going to skip looking up the duck.co user until I can talk with @jbarrett about it.

MariagraziaAlastra:
@jdorweiler Np, I've done that part. 
It's tested, but could you please test it too? Thanks!

jdorweiler:
Tested and lgtm.  Thanks for the help!

MariagraziaAlastra:
Np, thank you!

PR #931, opened 2015-09-09
Dev Pipeline - add activity details for each IA with a PR

MariagraziaAlastra:
Last comment is always shown, while last commit is only displayed if the user committed after the last comment was written.
Example: a staff member leaves a comment highlighting a bug in the PR, then the user fixes it, commits the changes and doesn't write a comment afterwards - in that case we show both the last commit and the last comment.

**"Show activity details" is checked**
screenshot-maria duckduckgo com 5001 2015-09-09 22-38-14

**These have the last commit**
screenshot-maria duckduckgo com 5001 2015-09-09 22-39-34

**"Show activity details" is unchecked**
screenshot-maria duckduckgo com 5001 2015-09-09 22-38-39

### DIFF
--- a/src/js/ia/Helpers.js
+++ b/src/js/ia/Helpers.js
@@ -1,6 +1,7 @@
 (function(env) {
     // Handlebars helpers for IA Pages
     
+    // Return elapsed time expressed as days from now (e.g. 5 days, 1 day, today)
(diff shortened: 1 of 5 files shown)


russellholt:
LGTM

PR #935, opened 2015-09-15
Only mail claimed ideas once.

jbarrett:
@russellholt @MariagraziaAlastra Not sure how I managed this, but this change removes ddgc-ia@ from the Cc field.


MariagraziaAlastra:
Thanks @jbarrett, looks good! :+1:

PR #937, opened 2015-09-16
IA Pages - add ability to ghost pages

MariagraziaAlastra:
I'm introducing a new dev milestone: "ghosted".
Pages with this milestone won't be fetched anywhere , except for inside ia/dev/deprecated, where you will be able to see them if you're an admin - just click on the "Show ghosted pages" checkbox on top. 
This is a temporary solution, since ghosted pages should be on a separate admin page.

TODO: change meta_id when the page is ghosted

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -179,6 +179,9 @@ sub deprecated_base :Chained('overview_base') :PathPart('deprecated') :CaptureAr
     $c->stash->{ia_page} = "IADeprecated";
     $c->stash->{title} = "Deprecated IA Pages";
    
+    $c->stash->{logged_in} = $c->user;
+    $c->stash->{is_admin} = $c->user? $c->user->admin : 0;
+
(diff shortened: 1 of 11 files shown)


MariagraziaAlastra:
Ok, now the page changes meta_id when it's ghosted, and it also uses a UUID to ensure the generated value it's unique.

E.g. if the previous meta_id was "airlines", after ghosting it it will be "airlines_ghosted_FF6ECF66-5C9A-11E5-84AD-25B28BD11E5D" (of course the UUID here is just an example).

This way we will be able to use again the former meta_id of the ghosted page, in this case "airlines", for another IA Page.

MariagraziaAlastra:
Before clicking on the checkbox, the ghosted pages are hidden:
screenshot-maria duckduckgo com 5001 2015-09-16 20-40-34

After clicking on the checkbox:
screenshot-maria duckduckgo com 5001 2015-09-16 20-42-12

I prefer having ghosted pages completely separated from deprecated pages, so this is the short-term visual solution I came up with. Ideally as I said they should be on another page, in some admin panel.

To restore a ghosted page just change its dev milestone to something other than "ghosted". The meta_id will have to be manually changed in that case.

russellholt:
LGTM :tv:

PR #940, opened 2015-09-17
Update composition of subscription types for less repetition.

jbarrett:


nilnilnil:
Is this WIP? Reviewer?

nilnilnil:
@jbarrett I guess I asked for it :smile:. What about @MariagraziaAlastra and friends?

jbarrett:
You may be confusing this with a project which has that sort of process and coverage... :smiling_imp: 

I would be obliged if you could, if you have the time.

Hey, we have test coverage for this one too! :)

nilnilnil:
@jbarrett Of course I'll give it a look! I'd like to get another pair of eyes on it too.

jbarrett:
@MariagraziaAlastra Feel free to reassign :smile:

jbarrett:
Thanks!

nilnilnil:
:

nilnilnil:
:+1: -- assigning to @MariagraziaAlastra for final say so. :smile:

lib/DDGC/Config/Subscriptions.pm
```diff
@@ -19,18 +19,27 @@ sub created_ia_page {
...
+    $self->_subscription(
+        'instant_answer',
+        'updated',
+        $meta,
+    );
+}
+
+sub ia_page {
+    my ( $self, $id ) = @_;
+    $self->updated_ia_page( { meta1 => $id } );
```
nilnilnil: I keep going back and forth on whether or not we should have `id` as an intrinsic field instead. I suppose that speaks to the fact we don't need to change it...

jbarrett: Yeah, I fumbled around with this too.

Entity / ID links are common enough in the project and work well enough, but I think I eschewed that here since we're not actually linking to entities / creating relationships. We are just describing the event parameters. It should be equally usable for events within and without the system.

*shrug* We're not so tied to anything that we can't change it.

MariagraziaAlastra:
@jbarrett LGTM! :+1:

PR #941, opened 2015-09-17
Ideas - add name field and use it for the IA Page name and meta_id

MariagraziaAlastra:
When a user opens a new idea thread, he will have the ability to choose a name for the Instant Answer (also any forum manager can change that value by editing the post), which is used as a name for the new IA Page instead of the title, and also as meta_id, after being formatted properly, if no other IA Page is using the same meta_id value.

The name will be visible in the thread if idea is unclaimed and thus there's no IA Page for it. 
Once it's created, though, it will be replaced by a link to it (while I was at it I moved it outside of the description).

Some screenshots:
screenshot-maria duckduckgo com 5001 2015-09-17 18-30-59

screenshot-maria duckduckgo com 5001 2015-09-17 18-29-50

screenshot-maria duckduckgo com 5001 2015-09-17 18-32-04

screenshot-maria duckduckgo com 5001 2015-09-17 18-32-24


MariagraziaAlastra:
@russellholt @jbarrett and anyone interested ^

lib/DDGC/Web/Controller/Ideas.pm
```diff
@@ -352,6 +363,20 @@ sub vote_view :Chained('vote') :PathPart('') :Args(0) {
...
 	$c->forward( $c->view('JSON') );
 }
+sub format_meta_id {
+    my( $id ) = @_;
+
+    # meta_id must be lowercase and without weird chars
+    $id = lc $id;
+    $id =~ s/[^a-z0-9]+/_/g;
+    $id =~ s/^[^a-zA-Z]+//;
+    $id =~ s/_$//;
```
russellholt: leading, trailing - good. might want to collapse multiple repeated ones also

MariagraziaAlastra: You can't have multiple repeated ones, because line 371 already turns multiple weird chars (underscore included) into a single underscore.

PR #942, opened 2015-09-17
Fix IA URLs in activity mail

jbarrett:
cc @MariagraziaAlastra Thanks!


MariagraziaAlastra:
Looks good, thanks @jbarrett!

PR #944, opened 2015-09-17
Add modified time to instant_answer

zachthompson:
Add `last_modified` to instant_answer table
Trigger to update `last_modified` on insert/update

### DIFF
--- a/sql/1.079/modified-time.sql
+++ b/sql/1.079/modified-time.sql
@@ -1,0 +1,15 @@
+alter table instant_answer add column last_modified timestamp with time zone not null default now();
+
+CREATE OR REPLACE FUNCTION update_last_modified()
+RETURNS TRIGGER AS $$
+BEGIN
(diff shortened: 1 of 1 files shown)


russellholt:
LGTM ...
@jdorweiler @jbarrett @MariagraziaAlastra

jbarrett:
I've been thinking about this since last night and, for various reasons, don't think we should go ahead with this change.

I should have an alternative later today (including whatever controller stuff is required to implement `Last-Modified`).

At the moment, all of our DDL can be derived from DBIC and is currently reasonably portable. While I would like to take much better advantage of postgres features (like full-text search), we're not quite equipped yet for having postgres do any advanced work. We're not (yet) equipped to deploy / migrate schemas from our own DDL.

I think it would be more expedient for now to deliver this the "DBIC-way" (as much as that fits in with our version of it).

Thanks!

jbarrett:
Thanks, @zachthompson

PR #946, opened 2015-09-22
Dev Pipeline - add actions

MariagraziaAlastra:
Now it's possible to select multiple IAs in the Dev Pipeline and change either the dev milestone or the type for all of them.
Still needs some visual tweak, but it's almost done!


MariagraziaAlastra:
For now the page is reloaded after saving - just trying to experiment quickly here.
At first the actions aren't displayed, they only become visible when at least one IA is selected. 
screenshot-maria duckduckgo com 5001 2015-09-22 22-29-37
screenshot-maria duckduckgo com 5001 2015-09-22 22-30-11

Issue #356, opened 2014-12-08
Make caching policy context sensitive.


jbarrett:
Currently there is a blanket expiry policy for all pages. Search result pages and public pages with comment forms should probably not be included.

Issue #947, opened 2015-09-23
More 304


jbarrett:
We should examine where we can short-circuit requests with 304 responses. Probably a lot of gains to be had here.

PR #948, opened 2015-09-24
Set Cache-Control more selectively.

jbarrett:
Setting no-cache in the root route handler (applying it to all requests) causes various annoyances and breakages. Let's stop doing it.


jbarrett:
This still sets no-cache for user settings pages, admin pages and moderation pages.

jbarrett:
Paves the way for #947 to work in browsers, Fixes #356

PR #951, opened 2015-09-24
Finishing Last-Modified implimentation

jbarrett:
I refactored this a little and fleshed it out, since I think it could be generally very useful. Ideally we could use something like [`Plack::Middleware::ConditionalGET`](https://metacpan.org/pod/Plack::Middleware::ConditionalGET) but since that only gets applied after our processing, we wouldn't get a huge time gain (just bandwidth) - processes would still manipulate the data, we just wouldn't send the result. Changes:

- `If-Modified-Since` test and header generation moved out to base controller class
- `->last_modified` moved out to base resultset class (this won't crash if the table has no updated column, but won't be accurate either - apply with care)
- Removing Content-Type etc. headers in 304 response
- Use string equality for if-modified test, since it's basically a tag and not a temporal consideration.
- Fix for IE (via `Plack::Middleware::ConditionalGET`)

This comes into its own alongside #948.

To use it, add it to your controller once you know which result or resultset class you're going to process like so:

```perl
$c->return_if_not_modified( $result->updated );
```

...or...

```perl
$c->return_if_not_modified( $resultset->last_updated );
```

### DIFF
--- a/lib/DDGC/DB/Base/ResultSet.pm
+++ b/lib/DDGC/DB/Base/ResultSet.pm
@@ -3,6 +3,9 @@ package DDGC::DB::Base::ResultSet;
 
 use Moose;
 use namespace::autoclean;
+use DateTime;
+use DateTime::Format::Pg;
+use Try::Tiny;
 
 extends 'DBIx::Class::ResultSet';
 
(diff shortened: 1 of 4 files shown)


jbarrett:
It should be noted that we can't use this in all instances, like on pages with forms + CSRF tokens for logged in users.

jbarrett:
Also, you don't specifically need result or rs class instances here, the parameter is a [`DateTime`](https://metacpan.org/pod/DateTime) instance so once you can derive that from what you're serving you can use it.

PR #953, opened 2015-09-25
Set status on dzil release to duckpan

zachthompson:
In order to better track the status of an IA after it's been merged, this PR attempts to:

1. Determine the IAs that have been updated in the release from the changelog.
2. Update release versions and statuses based on the type of change.

The release_version table tracks additions, updates, and deletions.

@jbarrett and @MariagraziaAlastra might be interested in these changes :hammer:


lib/DDGC/DB/ResultSet/ReleaseVersion.pm
```diff
@@ -0,0 +1,20 @@
+package DDGC::DB::ResultSet::ReleaseVersion;
+# ABSTRACT: Resultset class for tracking release versions 
+
+use Moose;
+extends 'DDGC::DB::Base::ResultSet';
+use namespace::autoclean;
+use DateTime;
+use DateTime::Format::Pg;
+
+sub last_modified {
```
jbarrett: This is in the base ResultSet class so shouldn't need to be defined here (that change just went in on Friday)

sql/1.079/modified-time.sql
```diff
@@ -0,0 +1,15 @@
+alter table instant_answer add column last_modified timestamp with time zone not null default now();
+
+CREATE OR REPLACE FUNCTION update_last_modified()
```
jbarrett: Do we still need this? I don't see it referenced elsewhere?

sql/x/update-release.sql
```diff
@@ -0,0 +1,10 @@
+alter table instant_answer add column release_version numeric(8,3);
```
jbarrett: What's the difference between this and `release_version.release_version`?

jbarrett:
:+1: Thanks, @zachthompson

lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -413,6 +413,10 @@ column updated => {
     set_on_update => 1,
 };
 
+column release_version => {
+    data_type => 'numeric',
+    is_nullable => 1,
+};
```
MariagraziaAlastra: @zachthompson Why not make a relation to the ReleaseVersion table? Just asking out of curiosity - I can definitely imagine it could be useful in the future if we want to see diffs and approve them for release directly on IA Pages, but of course for now it's not necessary!

MariagraziaAlastra:
Looks good, thanks @zachthompson!

lib/DDGC/DB/ResultSet/ReleaseVersion.pm
zachthompson: Will remove and test.

sql/1.079/modified-time.sql
zachthompson: Nope,  thought I removed all of that.

sql/x/update-release.sql
zachthompson: Nothing.

lib/DDGC/DB/Result/InstantAnswer.pm
zachthompson: It's a good idea.  There definitely is a relation.  I'll take a look at it.

zachthompson:
Think all of the suggestions have been addressed.  
released-ia

One additional note.  AFAIK we should be able to put a unique index on instant_answer.meta_id.  However there are dupe ids for the same meta_id.  I added delete statements for these to the sql for the foreign key.  If we want to do this manually, that's fine.

MariagraziaAlastra:
@zachthompson yeah that's what we want but need to figure out what's causing that bug... In the meantime another way to get rid of the dupes is to set the dev milestone to "ghosted".

Cool, I guess we could have the "released" ones showing up as coming soon in the index! Just a thought.

lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -425,6 +429,8 @@ many_to_many 'users', 'instant_answer_users', 'user';
 has_many 'instant_answer_topics', 'DDGC::DB::Result::InstantAnswer::Topics', 'instant_answer_id';
 many_to_many 'topics', 'instant_answer_topics', 'topic';
 
+has_many 'release_versions', 'DDGC::DB::Result::ReleaseVersion', 'instant_answer_id';
```
MariagraziaAlastra: :+1:

zachthompson:
I want to verify those dupes.  I'll also see if I can backfill the versions for as many existing IAs as possible.

zachthompson:
@MariagraziaAlastra like I said earlier, I think we're ready to review this.  It has several moving parts to be able to test which I can walk you through.

MariagraziaAlastra:
Hey @zachthompson!
I can tell you right here and right now that I like the deployment_state column idea more than the previous strategy, and yeah it would be cool if you could walk me through your changes. Thanks!

sql/1.084/01-update-release.sql
```diff
@@ -0,0 +1,17 @@
+BEGIN:
```
MariagraziaAlastra: Gives error here. Maybe get rid of the ":"? I think that's not necessary

zachthompson: Ah yes!

MariagraziaAlastra:
@zachthompson Tested and LGTM. Also love the code and the concept overall! Thank you! :+1:

PR #954, opened 2015-09-28
updated paths

zekiel:
updated paths because production currently has an extra "/stats" which causes 404 when clicking things at https://duck.co/admin/stats/


kablamo:
:+1: Thanks for fixing that!

PR #970, opened 2015-10-05
Developer / IA Notification updates.

jbarrett:
I meant to push this Friday to start discussing it. Anyway, so far there are a couple of changes here:

- We no longer generate "activity" for every update
- URLs have been "fixed" to use the current host - I found everything being 'duck.co' confusing

Since we don't (that I am aware of) have a ref to the current request there, doing reliable URL stuff from the model is a little funky. This fleshes out the `web_base` config option and adds a wrapper around [`URI`](https://metacpan.org/pod/URI) to generate full URLs for paths and request parameters.

Cc @MariagraziaAlastra

### DIFF
--- a/lib/DDGC/Config.pm
+++ b/lib/DDGC/Config.pm
@@ -72,7 +72,11 @@ sub rootdir {
 	return File::Spec->rel2abs( $dir );
 }
 
-has_conf web_base => DDGC_WEB_BASE => 'https://duck.co';
+has_conf web_base => DDGC_WEB_BASE => sub {
+	( $_[0]->is_live )
+		? 'https://duck.co'
+		: 'https://view.dukgo.com';
+};
 
 sub prosody_db_samplefile { File::Spec->rel2abs( File::Spec->catfile( dist_dir('DDGC'), 'ddgc.prosody.sqlite' ) ) }
 sub duckpan_cdh_template { File::Spec->rel2abs( File::Spec->catfile( dist_dir('DDGC'), 'perldoc', 'duckpan.html' ) ) }
--- a/lib/DDGC/DB/Result/InstantAnswer.pm
+++ b/lib/DDGC/DB/Result/InstantAnswer.pm
@@ -12,6 +12,13 @@ use JSON;
 table 'instant_answer';
 
 sub u { [ 'InstantAnswer', 'view', $_[0]->id ] }
+sub uri {
+    my ( $self, $params ) = @_;
+    $self->ddgc->uri_for(
+        sprintf( '/ia/view/%s', $self->meta_id ),
+        $params
+    );
+}
 
 column id => {
 	data_type => 'text',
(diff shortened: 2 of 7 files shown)


lib/DDGC/Config.pm
```diff
@@ -72,7 +72,11 @@ sub rootdir {
 	return File::Spec->rel2abs( $dir );
 }
 
-has_conf web_base => DDGC_WEB_BASE => 'https://duck.co';
+has_conf web_base => DDGC_WEB_BASE => sub {
+	( $_[0]->is_live )
+		? 'https://duck.co'
+		: 'https://view.dukgo.com';
```
MariagraziaAlastra: Nice, this way it'll be easier to test

lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -12,6 +12,13 @@ use JSON;
 table 'instant_answer';
 
 sub u { [ 'InstantAnswer', 'view', $_[0]->id ] }
+sub uri {
+    my ( $self, $params ) = @_;
+    $self->ddgc->uri_for(
+        sprintf( '/ia/view/%s', $self->meta_id ),
```
MariagraziaAlastra: :+1:

lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -433,11 +440,39 @@ after insert => sub {
...
+        description  => $description,
+    );
+}
+
+sub _generate_updates {
+    my ( $self, $update ) = @_;
+
+    while ( my ($column, $value) = each $update ) {
+
+        if ( $column eq 'dev_milestone' ) {
```
MariagraziaAlastra: So we generate updates only when changing dev milestone? Good one, that's the most important field change so far. 
In the future I'd like to see all changes to important fields made by another user (not the one getting the notification mail).

MariagraziaAlastra:
@jbarrett Looks good! :+1:

lib/DDGC/DB/Result/InstantAnswer.pm
jbarrett: Yeah, we could probably attach a `by_user` column to `activity_feed` which would contain `current_user` - simple to filter these entries out then, I think.

jbarrett: I just picked `dev_milestone` sort of arbitrarily (as the one which juat appears most important) - are there others we should include in this change?

MariagraziaAlastra: Sounds good

MariagraziaAlastra: The most important  (excluding dev milestone) are:
- name
- meta_id (it can change since it's editable)
- type
- perl module

For the purpose of this PR I think we're good with just dev milestone. Maybe meta_id too since it decides the URL

PR #973, opened 2015-10-05
_dev.scss: Don't modify the position anymore

jagtalon:
It's not aligning properly. There seems to be a problem with the dropdown on firefox but that's another issue. Works on Webkit browsers.

For @MariagraziaAlastra


MariagraziaAlastra:
@jagtalon Looks good, thanks!

PR #976, opened 2015-10-06
store all @mentions from last comment

jdorweiler:
Stores the last n @mentions from a comment.  
selection_034


MariagraziaAlastra:
@jdorweiler Thanks, looks good! :+1:

PR #985, opened 2015-10-15
_pipeline.scss: "F" and "L" content should be applied to :before element

jagtalon:
Labels for Fatheads and Longtails should now be correct (it was showing "N" before). 😦 For @MariagraziaAlastra.


MariagraziaAlastra:
Thanks @jagtalon!

PR #986, opened 2015-10-15
_pipeline.scss: Add .on-hold class.

jagtalon:
The .on-hold class sets opacity to 0.5 which signifies that an IA is, well, on hold.


MariagraziaAlastra:
Thanks @jagtalon!

PR #987, opened 2015-10-15
Dev Pipeline - use icons for filtering and add count

MariagraziaAlastra:
Also, now the filtering is only inclusive with the query box.
This means you can filter either by priority or missing info + use a query, like in the mockup.


MariagraziaAlastra:
screenshot-maria duckduckgo com 5001 2015-10-15 19-34-39

PR #988, opened 2015-10-15
_pipeline.scss: Add underline under the filter.

jagtalon:
Using `box-shadow` because using `border` takes up space. We don't want no space. 🚀


jagtalon:
screen shot 2015-10-15 at 2 01 24 pm

jagtalon:
Actually, wait, not yet done

MariagraziaAlastra:
Looking good, thanks @jagtalon! :+1:

PR #996, opened 2015-10-21
Style the actions sidebar

jagtalon:
For @MariagraziaAlastra. Currently a WIP.


jagtalon:
I think I got it close to the mockup in https://projects.invisionapp.com/share/ER482O582#/screens/102135940, but it definitely still needs another pass.

I need the following before continuing:

- Star
- Move
- The blue bar

MariagraziaAlastra:
:+1:

PR #997, opened 2015-10-21
check for status array

jdorweiler:
Need to check that obj exists otherwise this fails at `obj.length`


MariagraziaAlastra:
:+1:

PR #1001, opened 2015-10-23
Github::UserLink - migrate the data and make use of the stored usernames

MariagraziaAlastra:
This is blocking the dev pipeline redesign, so we need this done ASAP.
Assigning to @jbarrett for review, still WIP

### DIFF
--- a/lib/DDGC/DB/Result/GitHub/UserLink.pm
+++ b/lib/DDGC/DB/Result/GitHub/UserLink.pm
@@ -9,6 +9,6 @@ use namespace::autoclean;
 table 'github_user_link';
 
 primary_column users_id       => {data_type => 'bigint', is_nullable => 0};
-column data                   => {data_type => 'text',   is_nullable => 0, serializer_class => 'JSON'};
+column gh_data                   => {data_type => 'text',   is_nullable => 0, serializer_class => 'JSON'};
 
 1;
--- a/lib/DDGC/DB/Result/GitHub/UserLink.pm
+++ b/lib/DDGC/DB/Result/GitHub/UserLink.pm
@@ -8,8 +8,6 @@ use namespace::autoclean;
 
 table 'github_user_link';
 
-primary_column id             => {data_type => 'bigint', is_auto_increment => 1};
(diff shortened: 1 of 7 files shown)


MariagraziaAlastra:
@jbarrett I think it's ready for review!

sql/1.095/github_user.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
+    CREATE TABLE github_user_link (
+        "users_id" bigint NOT NULL,
+        "data" text NOT NULL
+    );
+
+    ALTER TABLE users ADD COLUMN "github_user" text;
+COMMIT;
```
jbarrett: This table already exists, we're just wrapping the table def.

jbarrett: Just noticing the users `ALTER` - we need to remove the CREATE TABLE block or this will fail.

MariagraziaAlastra: @jbarrett Ah, right, for some reason I thought it only existed on wing and not in production :)
Removed now!

lib/DDGC/DB/Result/GitHub/UserLink.pm
```diff
@@ -0,0 +1,14 @@
...
+
+use Moose;
+extends 'DDGC::DB::Base::Result';
+use DBIx::Class::Candy;
+use namespace::autoclean;
+
+table 'github_user_link';
+
+primary_column users_id       => {data_type => 'bigint', is_nullable => 0};
+column data                   => {data_type => 'text',   is_nullable => 0, serializer_class => 'JSON'};
```
jbarrett: My bad - this column is called `gh_data`

MariagraziaAlastra: @jbarrett Np, renamed!

jbarrett:
Thanks @MariagraziaAlastra!

Appears to be working, we can probably go ahead with this today.

MariagraziaAlastra:
Nice, thanks @jbarrett! I'll merge then :)

PR #1009, opened 2015-10-28
Dev Pipeline - sort by priority

MariagraziaAlastra:
This is just an experiment! I connected the TestOne account to one of the admins accounts on github, so whenever you see an IA with not so much elapsed time compared to the others on top, it means the admin was mentioned in the last PR comment.
Right now the parameters taken into account are:
- Last comment author (higher priority if non-admin)
- Elapsed time since last comment (again, more points if non-admin): the older, the higher the priority
- Elapsed time since last commit if made after last comment or if no comments
- User has been at mentioned in the last comment
- Tags (on hold: -100; high priority: +50)
- Has a PR (any IA with a PR is higher priority than IAs without one)
screenshot-maria duckduckgo com 5001 2015-10-28 12-41-32

### DIFF
--- a/src/js/ia/Helpers.js
+++ b/src/js/ia/Helpers.js
@@ -2,14 +2,15 @@
     // Handlebars helpers for IA Pages
     
     // Return elapsed time expressed as days from now (e.g. 5 days, 1 day, today)
-    Handlebars.registerHelper("timeago", function(date) {
+    Handlebars.registerHelper("timeago", function(date, full) {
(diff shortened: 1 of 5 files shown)


MariagraziaAlastra:
Sort by priority (default on page load):
screenshot-maria duckduckgo com 5001 2015-10-28 17-50-18

Sort by Recently Updated:
screenshot-maria duckduckgo com 5001 2015-10-28 17-50-51

PR #1020, opened 2015-10-30
Fix iconfont and hovering issues.

jagtalon:
For @MariagraziaAlastra.

- Iconfont was being annoying, so I used SVG instead.
- Links are now blue on hover.


MariagraziaAlastra:
:+1:

PR #1023, opened 2015-10-31
Simple status page

jbarrett:
Reports:

- Version
- DB availability
- Prosody availability


jkv:
@jbarrett I would add a timestamp to the page so you know the content you are getting is not cached.

jbarrett:
@jkv Good call!

Last commit adds mail-style data in UTC to page like so:

"Date : Mon, 2 Nov 2015 14:40:55 +0000"

jkv:
:+1:

PR #1026, opened 2015-11-02
Quash test errors from multiple declaration

jbarrett:
@MariagraziaAlastra `$server` is declared above in this sub, so just reused it here - presuming this is OK. Thanks!


MariagraziaAlastra:
@jbarrett Thanks, LGTM :+1:

PR #1027, opened 2015-11-02
Blockgroup column

zachthompson:
* Add blockgroup to instant_answer
* Populate blockgroup with appropriate values from both Blocks and defaults.
* Add new table instant_answer_blockgroup and populate with current groups
* Constrain `instant_answer.blockroup` with the new table


lib/DDGC/DB/Result/InstantAnswer.pm
```diff
@@ -449,6 +450,12 @@ column deployment_state => {
     is_nullable => 1,
 };
 
+column blockgroup => {
+	date_type => 'varchar',
+	size => 20,
+	is_nullable => 1
```
MariagraziaAlastra: @zachthompson If you want to retrieve this in the repo/all JSON endpoint you need to add 'for_endpt => 1' here.
Not sure if you wanted that though!

MariagraziaAlastra:
@zachthompson Looks good, I just left a comment in the code. Thanks!

MariagraziaAlastra:
@zachthompson Feel free to merge, looks good to me!

sql/1.096/dynamic-blocks.sql
```diff
@@ -0,0 +1,75 @@
...
+
+copy instant_answer_blockgroup (blockgroup) from stdin;
+goodie_exlusive
+spice_killresult
+spice_nonexclusive
+goodie
+spice
+fathead
+longtail
+\.
```
MariagraziaAlastra: @zachthompson I get this error: 
ERROR:  end-of-copy marker corrupt

zachthompson: Can you try again?  I added statements to revert any changes that might have been made as well so you don't need to drop/alter table.  Just `\i` the file.  I was able to repeatedly run it ok.

MariagraziaAlastra: Thanks!

sql/1.096/dynamic-blocks.sql
```diff
@@ -0,0 +1,78 @@
...
+
+
+alter table instant_answer
+    add column blockgroup varchar(20) references instant_answer_blockgroup (blockgroup);
+    
+update instant_answer set blockgroup = 'goodie' where repo = 'goodies';
```
MariagraziaAlastra: @zachthompson I think probably here I get this error:
ERROR:  insert or update on table "instant_answer" violates foreign key constraint "instant_answer_blockgroup_fkey"
DETAIL:  Key (blockgroup)=(0) is not present in table "instant_answer_blockgroup".

MariagraziaAlastra: @zachthompson I fixed the faulty data in my instance and now it works! I don't think it will fail in production, since it was just a test IA and the repo is always chosen using one of the default values in a dropdown.

zachthompson: Are we good to merge it then?

MariagraziaAlastra: Yup!

PR #1036, opened 2015-11-04
IE is showing some overlapping text.

jagtalon:
Before:
screen shot 2015-11-04 at 9 30 03 am

After:
screen shot 2015-11-04 at 9 50 01 am

Tested in Safari, Chrome, FF, and IE.


MariagraziaAlastra:
LGTM (lol another one of these emoji-named branches) :+1:

PR #1037, opened 2015-11-05
Hax : Preload IA index JSON in template

jbarrett:
@MariagraziaAlastra So this "works" but I've hit a wall or two.

- Encoding - wondering what's the best way to safely get the JSON string into the template, then into the js? The manual approach here is not robust ([raw strings, yich](https://github.com/duckduckgo/community-platform/blob/e8f8e1e/templates/instantanswer/index.tx#L2)) and has problems:

BBC IA

- I return the list sorted by name, so there should be no need for [this](https://github.com/duckduckgo/community-platform/blob/e8f8e1e/src/js/ia/IAIndex.js#L33) but when I remove it the index doesn't load - I'm guessing I misunderstand the purpose of this line.

I'd be interested in your impression of the difference in load times. Another experiment based on this code should be incoming later.

### DIFF
--- a/lib/DDGC/DB/ResultSet/InstantAnswer.pm
+++ b/lib/DDGC/DB/ResultSet/InstantAnswer.pm
@@ -5,4 +5,32 @@ use Moose;
 extends 'DDGC::DB::Base::ResultSet';
 use namespace::autoclean;
 
+sub ia_index_hri {
+    my ( $self, $limit, $last ) = @_;
+    my $ial = $self->search( {
+            'topic.name' => { '!=' => 'test' },
+            'me.dev_milestone' => { '=' => 'live'},
+        },
+        {
+            prefetch => { instant_answer_topics => 'topic' },
+            columns => [
+                qw/ name repo src_name
+                    dev_milestone description
+                    template id meta_id
+                /, ],
+            collapse => 1,
+        },
+    )->hri
+     ->order_by( 'me.name' );
+
+    $ial = $ial->rows( $limit ) if $limit;
+    $ial = $ial->search( {
+       'me.name' => { '>' => $last },
+    } ) if $last;
+
+    return $ial->all_ref;
+}
+
+
+
 1;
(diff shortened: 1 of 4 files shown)


MariagraziaAlastra:
@jbarrett Wow, thanks, I'll test this now!

You can remove the call to the sort function as long as you just call the refresh() function instead.
This line is why the index wasn't loading: https://github.com/duckduckgo/community-platform/blob/e8f8e1e/src/js/ia/IAIndex.js#L413

About the JSON string: yeah that doesn't look like a safe way to embed it... Maybe try passing the object to a function instead?

MariagraziaAlastra:
@jbarrett It loads instantly! I didn't have to wait at all! It was already there with all the data and... and... I might cry with joy :sob:

jbarrett:
Thanks for that @MariagraziaAlastra.

I changed the quote escaping in the JSON so it's still a raw value in the template... It "works" but I don't particularly like it.

I guess the next step would be to limit this initial set to some arbitrary number, 30 or 50 IAs, then retrieve the rest on load. The JSON endpoint now takes parameters:

@jagtalon Any idea on the best way to handle that?

/ia/json will now take a `last` parameter, so IAs after that in the sorted-by-name index should be start after that IA. It will also take a `limit` parameter if you want to do staggered loading on scroll or something, like so:

`/ia/json?last=Pokemon&limit=30`

Get the next 30 IAs after Pokemon.

jbarrett:
Thinking on this, I think the filtering would be pretty difficult without the upfront load - you'd spend a lot of time checking you had the full set for each filter.

MariagraziaAlastra:
Yeah @jbarrett we filter exclusively on the front-end, using CSS classes, which means it's probably not ideal loading just a limited set of IAs at the beginning.
Or whenever one of the filter options is selected, we could make a POST request to get all the remaining IAs? Not sure

jbarrett:
@MariagraziaAlastra Since this seems useful and I have pinned down reasonable behaviour, going to open a new PR with just the preload stuff.

MariagraziaAlastra:
Sounds good, thanks!

PR #1045, opened 2015-11-09
Return 1:0, not result instance for $user->is

jbarrett:
@jdorweiler @MariagraziaAlastra My bad, I thought we were just looking for truthiness from this test - didn't consider the result would be assigned for other stuff. This should resolve the useless Result instances finding their way into ghIssues.


MariagraziaAlastra:
@jbarrett Thanks, LGTM!

PR #1046, opened 2015-11-09
Preload IA index in template

jbarrett:
So following on from #1037, as a stop-gap performance measure this works well enough for now. It will be less feasible as the index grows and some day we will need to fragment loading across multiple requests. Until then...

Here are just the changes required to do the preloading.

@MariagraziaAlastra @russellholt @jagtalon

### DIFF
--- a/lib/DDGC/DB/ResultSet/InstantAnswer.pm
+++ b/lib/DDGC/DB/ResultSet/InstantAnswer.pm
@@ -8,7 +8,7 @@ use namespace::autoclean;
 sub ia_index_hri {
     my ( $self, $limit, $last ) = @_;
     my $ial = $self->search( {
-            'topic.name' => { '!=' => 'test' },
+            'topic.name' => [ { '!=' => 'test' }, undef ],
             'me.dev_milestone' => { '=' => 'live'},
         },
         {
--- a/lib/DDGC/DB/ResultSet/InstantAnswer.pm
+++ b/lib/DDGC/DB/ResultSet/InstantAnswer.pm
@@ -5,4 +5,30 @@ use Moose;
 extends 'DDGC::DB::Base::ResultSet';
 use namespace::autoclean;
 
+sub ia_index_hri {
+    my ( $self, $limit, $last ) = @_;
+    my $ial = $self->search( {
+            'topic.name' => { '!=' => 'test' },
+            'me.dev_milestone' => { '=' => 'live'},
+        },
+        {
+            prefetch => { instant_answer_topics => 'topic' },
+            columns => [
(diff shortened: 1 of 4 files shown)


MariagraziaAlastra:
LGTM, would be interested in a bit more background on what you did to make it faster! Thanks @jbarrett

jbarrett:
I think it's mostly an illusion - without the preload, we paint the frame and then the content pops in. This "feels" slow (to me anyway), even if it happens in less than 250ms. We offload this data transfer time to the server side template rendering instead. There are probably a few ms shaved off here or there by other stuff:

Other changes:
- Sort is done by postgres, not js
- JSON string is assigned as an object literal
    - This means no JSON.parse etc. is explicitly invoked, but I don't know if that makes much of a difference
    - I'm also not sure this is 100% robust - JSON should be a strict subset of object literal notation (I think!)
- JSON serialisation is done by [Cpanel::JSON::XS](https://metacpan.org/pod/Cpanel::JSON::XS) via [JSON::MaybeXS](https://metacpan.org/pod/JSON::MaybeXS)
    - ...though our Catalyst JSON View class should do that...
    - ...yet the IA controller still uses JSON in other places which will fall back to [JSON::PP](https://metacpan.org/pod/JSON::PP) in our case - slow!
    - ...but that should have no effect on the index JSON

Other considerations:
- Increased server load
    - This change increases server load since we no longer do an if-modified-since check - JSON is fetched and rendered every time!

MariagraziaAlastra:
I certainly looks better, and feels way faster :)
Although the server load issue should be carefully evaluated, I think. Can we afford to fetch JSON every time?

jbarrett:
We did it this way for a long time anyway. Even when we 304 the index JSON, it's still an extra request and still results in the pop-in (although it's a little faster than the 200 download of course). I think this is a decent tradeoff, once we can demonstrate it works like the existing code.

I think as long as these requests come from people and not automated processes we can handle it (which they should, since it's a request for the page).

Another thing I'm not sure on, I made the inline JSON literal a return from a function - does that object have to be compiled as soon as it is encountered in the markup or is it built when the function is called? I think doing the latter will make our time to first paint quicker without adversely effecting overall render time, but I'm not really sure how to effectively benchmark that.

MariagraziaAlastra:
Alright then, I'm on board with these changes so far.

Not sure about when the JSON gets built actually... I would think it's the latter though

jbarrett:
@MariagraziaAlastra Going to assign to you, so. Thanks!

MariagraziaAlastra:
Looks good, merging. Thanks @jbarrett! :+1:

PR #1047, opened 2015-11-10
Script to mirror IA repo_all JSON on S3

jbarrett:
- Only mirrors from duck.co if modified
- Uploads to S3 on curl success (includes 304 response)
- Nothing ties this to being run on duck.co
- Should be quiet running unless error occurs - cron suitable

@jkv @zachthompson

### DIFF
--- a/script/ddgc_upload_metadata.sh
+++ b/script/ddgc_upload_metadata.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+[ "$DDGC_REPO_JSON_OUT" == "" ]   && \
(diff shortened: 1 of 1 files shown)


jkv:
:+1: LGTM

@jbarrett, I changed the s3cmd into a single statement. I verified this works from the production instance as well.

jbarrett:
@jkv Super, thanks! Merge when ready, will ping you on release.

jkv:
@jbarrett I removed the curl and switched to use `LWP::Simple::mirror` so it would respect the 304 returned from duck.co. The script will only download the full JSON if it was modified from duck.co and it will only upload to S3 if the local JSON file was modified.

Merging it in now.

PR #1050, opened 2015-11-11
User is subscribed to IA check

jbarrett:


jbarrett:
@MariagraziaAlastra I think the functions here should be enough to implement basic follow / unfollow on IAs. Let me know if something else is required. Thanks

MariagraziaAlastra:
@jbarrett LGTM, thanks!

PR #1052, opened 2015-11-12
Fix annoying missing sessions dir

kablamo:
This has been driving me up the wall


jbarrett:
Thanks @kablamo - we should probably switch this out for a FastMmap or Memcache store at some stage... though maybe files are OK in development.

PR #1056, opened 2015-11-12
allow forcing an upload to s3

jkv:
Add an option to specify a parameter so it will force the upload to s3.

cc: @jbarrett


script/ddgc_upload_metadata.sh
```diff
@@ -1,11 +1,14 @@
 #!/usr/bin/env bash
 
+FORCE_UPLOAD=0
+[ -n "$1" ] && \
```
jbarrett: So this can be anything? I vote for --frank-zappa

jbarrett:
:+1:

PR #1057, opened 2015-11-12
IA Pages - edit blockgroup and deployment_state

MariagraziaAlastra:
Still need to add deployment_state!
Blockgroup editing and committing (only available to admins):
blockgroup_edit

screenshot-maria-old duckduckgo com 5001 2015-11-12 20-05-23


MariagraziaAlastra:
Added deployment_state:
deployment_state
screenshot-maria-old duckduckgo com 5001 2015-11-13 14-48-48

PR #1058, opened 2015-11-12
More design fixes

jagtalon:
- Set base font-size to 13px


MariagraziaAlastra:
@jagtalon Thanks! :+1:

jagtalon:
Thanks!!!

PR #1059, opened 2015-11-13
Cache deps on travis

zachthompson:
Use travis' caching to greatly reduce build times.  Will cache Perl, metadata, etc., and update the cache when they update.  This branch's build went from 19+ minutes to ~3 using the cache.


jbarrett:
:+1: Sweet, thanks @zachthompson

PR #1061, opened 2015-11-13
Fixing blog test...

jbarrett:
This PR is really preparing us to start using a version of Dancer2 which does not spit out **DEPRECATED** warnings for calling plugins within plugins.

v0.163000 has a bug which causes Plack to barf when using forward with the JSON serialiser. This is resolved in https://github.com/PerlDancer/Dancer2/pull/1040 but is not yet released.

In summary:

- This should fix current Travis CI errors
- This change is not harmful to the existing deployed version (that I know of)
- If we start using Dancer2 0.163000 to get rid of deprecation errors, this is useful
- The hook will not be harmful for versions **after** 0.163000 but should be removed

@MariagraziaAlastra @zachthompson Thanks!

### DIFF
--- a/lib/Dancer2/Plugin/DDGC/Service.pm
+++ b/lib/Dancer2/Plugin/DDGC/Service.pm
@@ -46,6 +46,23 @@ on_plugin_import {
             },
         )
     );
+
+    # https://github.com/PerlDancer/Dancer2/pull/1040
+    #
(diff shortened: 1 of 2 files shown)


MariagraziaAlastra:
Cool, thanks @jbarrett! :+1:

PR #1062, opened 2015-11-13
Gruntfile.js: Move the concatenation of the Charts library after Uglifying

jagtalon:


jagtalon:
For @jbarrett @MariagraziaAlastra

PR #1064, opened 2015-11-13
IAPage.js - comment out unused code, momentarily

MariagraziaAlastra:
Since we don't have traffic data in this release


jbarrett:
:+1: Thanks!

PR #1066, opened 2015-11-14
change ddg attribution name

jdorweiler:
Change ddg attribution name from "DDG Team" to "DuckDuckGo".  The sql only fixes IAs where the only dev is DDG Team.  There are a few with multiple devs that I'll have to fix by hand.  I need to wait for the  change to InstantAnswer.pm for that though.


MariagraziaAlastra:
:+1:

jdorweiler:
:cactus:

PR #1070, opened 2015-11-18
Populate dev IA data from /ia/repo/all/json

jbarrett:
@MariagraziaAlastra I would appreciate if you tried this out, let me know of any problems.

This does UPSERTs rather than nuking and rebuilding so you should be run it safely as often as required.

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -93,10 +93,13 @@ sub iarepo_json :Chained('iarepo') :PathPart('json') :Args(0) {
         ( $repo ne 'all' )
             ? ( repo => $repo )
             : (),
-        -or => [{dev_milestone => 'live'},
-        {dev_milestone => 'development'},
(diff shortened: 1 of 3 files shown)


MariagraziaAlastra:
@jbarrett So fast!
Anyway it did work without any problem, I'm still looking around to find any issue. 
Since this script uses the ia/repo/all/json endpoint, of course it won't fetch the IAs in planning, deprecated and ghosted dev milestones, but it's probably not important.
We can probably populate the db with IAs in planning by running script/ghIssues.pl anyway

jbarrett:
That last commit will allow you to ignore the milestones filter. It appears to work for me. Of course, it won't work for this script until this goes live.

MariagraziaAlastra:
Cool, thanks @jbarrett!

jbarrett:
@MariagraziaAlastra Hoppy to go ahead with this?

MariagraziaAlastra:
@jbarrett  Sure, works for me! :+1:

PR #1071, opened 2015-11-18
New IA flow prototype

MariagraziaAlastra:
Not yet functional:
screenshot-maria-old duckduckgo com 5001 2015-11-18 21-38-13


jbarrett:
Cool! Are we going to be directing New Idea to this or is it early days?

MariagraziaAlastra:
I think the "New Instant Answer" button on ddh should redirect to this, if we agree it works and it's what we want :)

PR #1072, opened 2015-11-19
Give sort a default value - absence commonly throwing exceptions

jbarrett:
@tagawa Just throwing 99 into sorting by default - can always be changed later.


tagawa:
Thanks.

PR #1074, opened 2015-11-20
Gruntfile.js: Cleanup

jagtalon:
There are concat tasks common between release and local builds. This prevents that repetition.


jagtalon:
This also prevents `grunt watch` from creating release files.

MariagraziaAlastra:
:+1:

PR #1075, opened 2015-11-20
Add new animation for the screenshot loader.

jagtalon:
- New retina-ified asset
- Uses CSS animation instead of GIFs for alpha transparency


MariagraziaAlastra:
@jagtalon Looks good, thanks! :+1:

PR #1080, opened 2015-11-20
Remove old static files

MariagraziaAlastra:
@jbarrett I kept the latest five as previously agreed :)


jbarrett:
I think we don't even need the last 5 - they'll be in the previous release if we need to roll back.

Can we make this part of grunt release?

MariagraziaAlastra:
@jbarrett Sure, looking into it!

PR #1082, opened 2015-11-23
IA Page: Don't require a test machine because we default to 'beta' an…

jagtalon:
…yway.

For @MariagraziaAlastra CC @moollaza


jagtalon:
screen shot 2015-11-23 at 3 29 34 pm

MariagraziaAlastra:
@jagtalon Thanks, LGTM! :+1:

PR #1085, opened 2015-11-25
Change foreign key on release_version

zachthompson:
The instant_answer.id column doesn't get updated, making it more suitable for the foreign key on release_version.


MariagraziaAlastra:
@zachthompson Looks good, thanks! :+1:

PR #1088, opened 2015-11-25
keep all issues from ghIssues

jdorweiler:
I changed a few things.  First, stop deleting the table when we ghIssues runs.  This way we will retain all the issues so we can link old PRs. 
Second, I added in the `since` filter when we request issues.  This will give us any issue that has been updated in that time frame which means we only really need to request a days worth of data.


MariagraziaAlastra:
@jdorweiler I first ran this script a couple of days ago and everything looked alright. Now I ran it again and everything is still good.
It actually works, e.g. the PR for this IA was merged but I still have it on my instance:
screenshot-maria-old duckduckgo com 5001 2015-11-30 21-00-58

Also, now it runs way faster.
If you agree I'd like to merge this today. Thanks!

jdorweiler:
@MariagraziaAlastra  sounds good to me!

MariagraziaAlastra:
@jdorweiler Well then, thanks for working on this! LGTM :+1:

PR #1091, opened 2015-11-26
[WIP] Improve speed of compiling SCSS files.

jagtalon:
In preparation for the improvements that we need to make for GQ&H (https://projects.invisionapp.com/share/5T52AX88Q0), I'd like to make a small addition to grunt to improve speeds. Currently, I'd have to wait for 6 seconds for `grunt watch` to build all the CSS files, and it starts to get annoying after that.

It's a WIP since I have yet to test if everything is working as before, but preliminary results on my machine have cut down the time from 6s to 1s. Here's what had to be done for the speedup:

1. Remove unnecessary compiling of Handlebars and JS files when only the CSS changed. (cut down time from 6s to 4s)
2.  Move to libsass. (cut down from 4s to 1s)

screen_shot_2015-11-26_at_3_36_44_pm
screen_shot_2015-11-26_at_3_51_02_pm
screen_shot_2015-11-26_at_4_11_00_pm

Testing:
- [x] `grunt watch` for SCSS files
- [x] `grunt watch` for JS files
- [x] `grunt watch` for Handlebars


templates/scripts.tx
```diff
@@ -18,3 +18,5 @@
 	<script src="/static/js/highlight.pack.js" type="text/javascript"></script>
 	<script>hljs.initHighlightingOnLoad();</script>
 : }
+
+<script src="//jag.duckduckgo.com:5000/livereload.js"></script>
```
jagtalon: @jbarrett @MariagraziaAlastra do you know how to add this for testing only? I'd like the browser to livereload only when developing: https://www.youtube.com/watch?v=X6wqhHD-P7M

jbarrett: Yep, we [stash a variable, `is_dev` in the root controller](https://github.com/duckduckgo/community-platform/blob/3e5a200/lib/DDGC/Web/Controller/Root.pm#L77) so you should be able to:

```
: if $is_dev {
    <script src="..."></script>
: }
```

Should the source for this script just be '/livereload.js'?

jagtalon: @jbarrett thank you!

jagtalon: It's a different port number so I need to specify a hostname. @jbarrett is there a variable that I can use to get the hostname?

jbarrett: Is the CSS on the same instance (port) as the one rendering the template?

MariagraziaAlastra:
@jagtalon Did you perhaps commit and push changes from other branches you were on, here? It looks like you added some SCSS changes in here as well

jagtalon:
@MariagraziaAlastra yeah I apologize for the mess I started doing the dev touchups in here as well since it was so nice to have something running quickly. I'll find a way to separate the two

MariagraziaAlastra:
@jagtalon Alright, cool, np!

PR #1095, opened 2015-11-30
Ideas Forum - remove claiming process

MariagraziaAlastra:
I left the db stuff in there, but everything else should be gone.
Erased from existence :christmas_tree:


jbarrett:
:+1: Thanks!

PR #1096, opened 2015-11-30
new_ia.tx - change fields order

MariagraziaAlastra:
//cc @russellholt @jagtalon description and name come before example query now:
screenshot-maria-old duckduckgo com 5001 2015-11-30 21-21-31


jagtalon:
@MariagraziaAlastra Can we simplify to "What does your Instant Answer do?" I'm not sure if I understand what "target" means completely.

MariagraziaAlastra:
@jagtalon Done, thanks for the suggestion! Since it's a small change I'm merging now, but feel free to let me know if you have any more feedback on this :)

PR #1100, opened 2015-12-03
Make ->admin act like an accessor

jbarrett:
Some stuff (notable dev database deploy) still uses `admin` as an accessor. This change allows stuff like `$user->admin(1)` to continue to work.

### DIFF
--- a/lib/DDGC/DB/Result/User.pm
+++ b/lib/DDGC/DB/Result/User.pm
@@ -242,7 +242,13 @@ sub unsubscribe_all_notifications {
 sub db { return shift; }
 
 sub translation_manager { shift->is('translation_manager') }
-sub admin { shift->is('admin') }
+sub admin {
+	my ( $self, $set ) = @_;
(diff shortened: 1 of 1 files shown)


MariagraziaAlastra:
@jbarrett Thanks! What would be a good way to test this? Apart from checking that nothing breaks across the site :)

jbarrett:
This is intended to fix ddgc_deploy_dev - Users like TestOne weren't being created as admins.

MariagraziaAlastra:
Cool, I had this issue recently! I'm killing the environment and then I'll build it again, see if I can get my users correctly set this time :)

jbarrett:
Thanks @MariagraziaAlastra - sorry the test is so destructive. At least we have a working IA import script now... once you can remember the name of it.

MariagraziaAlastra:
@jbarrett np at all!
The users are correctly set, thanks! LGTM :+1:

PR #1104, opened 2015-12-03
Wrapping up /wear

jbarrett:


jbarrett:
This has an external dependency : https://github.com/duckduckgo/duckduckgo-publisher/pull/216

...which should go live before this PR.

PR #1108, opened 2015-12-04
Dev Pipeline - add beta filter

MariagraziaAlastra:
//cc @jagtalon @jdorweiler 
screenshot-maria-old duckduckgo com 5001 2015-12-04 16-57-58

### DIFF
--- a/src/js/ia/Helpers.js
+++ b/src/js/ia/Helpers.js
@@ -129,6 +129,16 @@
         }
     });
 
+    // Check if val1 matches val2 using a regex
+    Handlebars.registerHelper('match', function(val1, val2, options) {
+       val2 = new RegExp(val2);
+       if (val1.match(val2)) {
+           return options.fn(this);
+       } else {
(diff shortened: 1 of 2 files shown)


src/js/ia/Helpers.js
```diff
@@ -129,6 +129,16 @@
         }
     });
 
+    // Check if val1 matches val2 using a regex
+    Handlebars.registerHelper('match', function(val1, val2, options) {
+       val2 = new RegExp(val2);
```
jagtalon: Isn't it always just the string 'success' if it's successful? Maybe just do `===`?

MariagraziaAlastra: Nope, it can also be "success: [some other stuff]", it's happened. I'll have to change the sidebar code as well in another PR

jagtalon: ok!

jagtalon:
:+1:

PR #1114, opened 2015-12-08
Login / Reg form changes

jbarrett:
- Stop using randomised field names
- Stop using captcha
- Start using CSRF token

Other potential changes coming in separately - I've made the mistake of trying to conflate too many changes in one PR before... One assumption winds up being invalid, all the other work has to go.

### DIFF
--- a/lib/DDGC/Web/Controller/My.pm
+++ b/lib/DDGC/Web/Controller/My.pm
@@ -76,14 +76,13 @@ sub login :Chained('logged_out') :Args(0) {
 
 	my $last_url = $c->session->{last_url};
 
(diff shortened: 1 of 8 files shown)


MariagraziaAlastra:
@jbarrett After I sign up, when I try to login it says "Login failed. Your username or password is incorrect."

jbarrett:
Bet it works in Private Browsing mode.

MariagraziaAlastra:
@jbarrett Oh yeah it does! I forgot about this private browsing mode thing, although I remember encountering it previously too.
Anyway, looks good to me! :+1:

jbarrett:
Thanks @MariagraziaAlastra - problem is the browser auto-filling hidden fields if you have it store passwords and such.

PR #1118, opened 2015-12-09
Endpoint to give a count of current DDH slack users.

jbarrett:


jbarrett:
This is not required.

PR #1119, opened 2015-12-09
only check and update file list for recently updated PRs

jdorweiler:
We're getting issues for anything that has recently been updated.   This means we no longer need to check and update the file list for every PR.

Cuts the number of api requests down by a lot.  If you want to see api requests add this env variable before running. `export NG_DEBUG=1`


MariagraziaAlastra:
@jdorweiler Awesome, thanks! :+1:

PR #1120, opened 2015-12-09
IA Overview - show IAs recently installed on beta

MariagraziaAlastra:
//cc @jagtalon 
screenshot-maria duckduckgo com 5001 2015-12-09 21-08-04


jagtalon:
@MariagraziaAlastra looks like the time isn't being parsed properly:

screen shot 2015-12-10 at 11 11 46 am

It says `347d ago` even though in the JSON it was only a month ago.

MariagraziaAlastra:
@jagtalon fixed

MariagraziaAlastra:
Since it should be ok now I'm merging, so we can test it on staging. Thanks!

PR #1126, opened 2015-12-14
dev_pipeline.tx - add again html element used for detecting permissions

MariagraziaAlastra:
//cc @jagtalon This brings back multiselect on the pipeline


jagtalon:
@MariagraziaAlastra oh got it I didn't know that that element was there for a reason

PR #1127, opened 2015-12-15
Have ddgc_upload_metadata's force upload option also force re-get

jbarrett:
Because I can't be trusted to tell you what's new, let's make the "force upload" option also re-fetch the data.

This is probably useful as an option to run on the shell when things don't look right on release.

Cc: @bsstoner @russellholt


jkv:
Looks good @jbarrett but I also added a check so the `rm` only happens if the file exists. Merge whenever you are ready.

jbarrett:
Thanks @jkv

PR #1130, opened 2015-12-15
Helpers: Fix faulty timeago helper.

jagtalon:
Two things that's faulty with the `timeago` helper:

1. `.replace` method should be a regex in the first parameter and not a string.
2. We can't have implicit variables in the handlebars helpers. ~~Perl idioms seem to have seeped into our JavaScript! `0` isn't the same as `false`. `0 ? 1 : 2` will always result to `2`.~~

Please review, @MariagraziaAlastra.

### DIFF
--- a/src/js/ia/Helpers.js
+++ b/src/js/ia/Helpers.js
@@ -30,7 +30,7 @@
         var timestring = full? " days ago" : "d";
         if (date) {
             // expected date format: YYYY-MM-DDTHH:mm:ssZ e.g. 2011-04-22T13:33:48Z
-            date = date.replace("/T.*Z/", "");
+            date = date.replace(/T.*Z/, "");
             date = moment.utc(date, format);
             
             var elapsed = parseInt(moment().diff(date, "days", true));
--- a/src/templates/dev_pipeline.handlebars
+++ b/src/templates/dev_pipeline.handlebars
@@ -142,12 +142,12 @@
                       {{#exists pr 'issue_id'}}
                         {{#eq pr.status 'open'}}
                               <a href="https://github.com/duckduckgo/zeroclickinfo-{{repo}}/issues/{{pr.issue_id}}" class="one-line" id="pr" title="{{format_time last_update}}">
-                                {{timeago last_update 0}}
+                                {{timeago last_update false false}}
                               </a>
                           {{else}}
                               {{#if last_update}}<i class="{{#eq pr.status 'merged'}} platform-pr-merged {{else}} platform-pr-closed {{/eq}}"></i>{{/if}}
                               <span class="no-pr" title="{{format_time last_update}}">
-                                {{timeago last_update 0}}
+                                {{timeago last_update false false}}
                               </span>
                           {{/eq}}
                       {{/exists}}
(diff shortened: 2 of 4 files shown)


src/js/ia/Helpers.js
```diff
@@ -30,7 +30,7 @@
         var timestring = full? " days ago" : "d";
         if (date) {
             // expected date format: YYYY-MM-DDTHH:mm:ssZ e.g. 2011-04-22T13:33:48Z
-            date = date.replace("/T.*Z/", "");
+            date = date.replace(/T.*Z/, "");
```
jdorweiler: Can momentjs use the date string as-is?  I found this in the docs http://momentjs.com/docs/#/parsing/special-formats/

jagtalon: @jdorweiler Yes I was thinking that too! I think this is a relic of _before_ we used momentjs. I could change it to that instead if we're not in a hurry

src/templates/dev_pipeline.handlebars
```diff
@@ -142,12 +142,12 @@
                       {{#exists pr 'issue_id'}}
                         {{#eq pr.status 'open'}}
                               <a href="https://github.com/duckduckgo/zeroclickinfo-{{repo}}/issues/{{pr.issue_id}}" class="one-line" id="pr" title="{{format_time last_update}}">
-                                {{timeago last_update 0}}
```
MariagraziaAlastra: @jagtalon You're not even supposed to use 0 here. It should just have no arguments. 
Options are the elements to render after the helper, and they never get passed, so it's not a problem to not pass a value for the from_beta value. Look at the overview and issues handlebars.

src/templates/dev_pipeline.handlebars
```diff
@@ -142,12 +142,12 @@
...
+                                {{timeago last_update false false}}
```
MariagraziaAlastra: false false? What? See previous comment

src/js/ia/Helpers.js
MariagraziaAlastra: @jagtalon @jdorweiler The reason why we're stripping the hours, minutes and seconds is that this way it's easier to get an int number (just days) as difference, when calculating the elapsed time.
There's no reason to change this part.

MariagraziaAlastra:
@jagtalon That "from_beta" variable is a recent addition to our code. It's supposed to be null (aka you shouldn't pass it at all) when the timestamp doesn't come from the beta server.
It was a way to just change the code inside the overview handlebars in the "Recently on beta" section without having to change it everywhere else as well.

jagtalon:
@MariagraziaAlastra oh ok lemme try it! Maybe I misunderstood

jagtalon:
Hey, @MariagraziaAlastra! I tested it, and I've verified that we have to have explicit arguments, e.g., `{{timeago last_update false false}}` instead of `{{timeago last_update}}` only, because an object is always passed as a parameter. Here's a quick demo (you can open [this page](https://s.codepen.io/jagtalon/debug/66536f6334d0aed68bce47559d1634a3) which has HandlebarsJS loaded, and you can type these in the console):

screen shot 2015-12-16 at 10 35 05 am

So in the first line I created a helper function that accepts one argument. I invoke that function using `{{test}}` and you'll see that the parameter `hi` is not `null`, but rather an object that is always passed as the last argument. So in our case, running `{{timeago last_update}}` on its own messes up our logic in the Handlebars helper function which causes the wrong dates to come up:

screen shot 2015-12-16 at 10 42 07 am

I hope that explains the use of `{{timeago last_update false false}}`!

MariagraziaAlastra:
LGTM, thanks @jagtalon

PR #1131, opened 2015-12-17
use try on beta requests

jdorweiler:
Just in case we don't get a response back.

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -223,7 +223,11 @@ sub dev_pipeline_json :Chained('dev_pipeline_base') :PathPart('json') :Args(0) {
     my $asana_server = "http://beta.duckduckgo.com/install?asana&ia=everything";
 
     my $result = asana_req('', $asana_server);
-    $result = $result->decoded_content ? from_json($result->decoded_content) : undef;
+    try{
+        $result = $result->decoded_content ? from_json($result->decoded_content) : undef;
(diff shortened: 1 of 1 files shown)


lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -934,7 +934,11 @@ sub asana_req {
     $req->header("x-hub-signature" => $header_data);
     $req->content($json_data);
 
-    my $result = $ua->request($req);
+    my $result;
+    try{
+        $result = $ua->request($req);
```
jbarrett: We already have a LWP::UA instance available in `$c->d->http`, no need to instantiate new ones.

jbarrett:
The response should probably be cached? See lines 49 and 58 [here](https://github.com/duckduckgo/community-platform/blob/e014113/lib/DDGC/Web/Controller/Duckduckhack.pm) for examples:

jdorweiler:
@jbarrett thanks. I'll make a task to add caching and `$c->d->http` later on unless you think it's needed now?

jdorweiler:
Tested it by switching from `beta.duckduckgo.com` to some other server that doesn't have the installed.json file.

jbarrett:
Not urgent.

To be clear, the issue was the JSON parser choking on HTTP content - I think you should be checking `$response->is_success` before doing anything with the content?

We should also make a task to stop using the plain JSON module in favour of [JSON::MaybeXS](https://metacpan.org/pod/JSON::MaybeXS)

PR #1133, opened 2015-12-17
New IA wizard - login and signup modals

MariagraziaAlastra:
Still need to automatically log in (I want to do that from the frontend, so we don't lose the new IA data) after signing up and I also need to handle errors better.
Almost done, though.
@jbarrett Feel free to take a look at the code already :) I made different endpoints for both login and signup from ia/new_ia.

### DIFF
--- a/src/js/ia/IAPageNew.js
+++ b/src/js/ia/IAPageNew.js
@@ -52,8 +52,18 @@
                 var username = $(prefix + "username-input").val();
                 var pwd = $(prefix + "pwd-input").val();
                 
+                $(".error-msg").addClass("hide");
+                
                 if (username.length && (pwd.length >= 8)) {
                     login(username, pwd);
+                } else {
+                   if (!username.length) {
+                      $("#login-username-error").removeClass("hide");
+                   }
+
+                   if (pwd.length < 8) {
+                      $("#login-pwd-error").removeClass("hide");
+                   }
                 }
             });
 
@@ -62,6 +72,8 @@
                 var username = $(prefix + "username-input").val();
                 var pwd = $(prefix + "pwd-input").val();
                 var email = $(prefix + "email-input").val();
+                
+                $(".error-msg").addClass("hide");
                 
                 if (username.length && (pwd.length >= 8)) {
 		    var req = $.post("/my/register_from_ia_wizard", {
@@ -74,8 +86,18 @@
 		        if (data && data.result) {
 		    	    $("#signup-bg, #signup-form").addClass("hide");
                             login(username, pwd);
-		        }
+		        } else {
+                            $("#signup-undef-error").removeClass("hide");
+                        }
 		    });
+                } else {
+                    if (!username.length) {
+                        $("#signup-username-error").removeClass("hide");
+                    }
+
+                    if (pwd.length < 8) {
+                        $("#signup-pwd-error").removeClass("hide");
+                    }
                 }
             });
             
@@ -99,7 +121,9 @@
 	                $("#login-bg, #login-form").addClass("hide");
 	                var data = getData();
 	                create_ia(data);
-	            }
+	            } else {
+                      $("#login-undef-error").removeClass("hide");
+                    }
 	        });
             }
 
--- a/src/js/ia/IAPageNew.js
+++ b/src/js/ia/IAPageNew.js
@@ -9,6 +9,7 @@
         init: function() {
             // console.log("IAPageNew init()");
             var page_new = this;
+            var username = $(".user-name").text();
                 
             // 100% width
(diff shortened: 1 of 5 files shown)


MariagraziaAlastra:
//cc @russellholt @jagtalon @jbarrett DONE!

The POST requests are sent using ajax, and I made two new endpoints, one for login and another one for the signup, specifically for this page, since we don't want any of the default behaviour here.
I also had to remove the unsetting of the action token after the signup to avoid crashing the site when performing the login immediately afterwards.

So this is how it works, for an unregistered user: after filling in the fields and clicking on "Create Instant Answer Page", the signup modal shows up. You can open the login modal instead by clicking on "sign in" inside this modal.
After signing up the user is automatically logged in, then the new IA Page is created and the user is redirected to it, with the next steps infobox open by default.
And now, screenshots! 
Signup:
screenshot-maria duckduckgo com 5001 2015-12-18 13-55-06

Login:
screenshot-maria duckduckgo com 5001 2015-12-18 13-55-22

Some basic input validation handling (we can do much more but this is enough for now, imho):
screenshot-maria duckduckgo com 5001 2015-12-18 13-58-34

russellholt:
@MariagraziaAlastra is this new code (perl) for user account setup and authentication?

MariagraziaAlastra:
@russellholt The user account setup and authentication code is the same, I pulled it off the default endpoints and made two subs that are called both by them and by the new endpoints. What's different for ia/new_ia the input validation is done on the frontend and the login/signup endpoints don't redirect anywhere.

templates/instantanswer/new_ia.tx
```diff
@@ -21,6 +21,52 @@
...
     
+    <div id="signup-bg" class="modal-bg hide"></div>
+    <div id="signup-form" class="hide modal-wrapper">
+        <div>
+            <h3 class="label left"> Create an Account</h3>
+            <div class="right" id="signup-cancel">
+                 <i class="ddgsi ddgsi-close"></i>
+            </div>
+            <p class="clearfix first-p">Thanks - but you'll need to create an account (or <a class="login_newia" href="#">sign in</a>) before you can proceed. Don't worry, it will be quick!</p>
+            <p class="error-msg hide" id="signup-undef-error">An error occurred. Please try again</p>
```
jbarrett: Can we get way more specific here? As far as I can tell I am doing everything right, but am getting this error.

jbarrett:
Can we make the return key submit login and reg forms?

src/js/ia/IAPageNew.js
```diff
@@ -41,8 +41,93 @@
...
+            });
+
+            $("#login-save").click(function(evt) {
+                var prefix = "#login-";
+                var username = $(prefix + "username-input").val();
+                var pwd = $(prefix + "pwd-input").val();
+                
+                $(".error-msg").addClass("hide");
+                
+                if (username.length && (pwd.length >= 8)) {
```
jbarrett: We shouldn't so any password validation on login (or, arguably, on registration... but people complain when we don't) - that's what the auth layer is for.

Username validation needs to be the same as the [existing regex](https://github.com/duckduckgo/community-platform/blob/755037f/lib/DDGC/Web/Controller/My.pm#L716-L719) and include the same feedback on the form.

The username forms part of a XMPP JID. I'm not sure what the valid set is for that, but the existing validation has worked out so far.

jagtalon:
Will check in a bit!

MariagraziaAlastra:
@jbarrett sure, I can make the return key submit, too!
Try filling in the email field using a valid address format, see if I left some backend email validation in

jbarrett:
I haven't been submitting email addresses :)

jbarrett:
Besides notes above, I am happy to stage this in principle since it does not disrupt existing login / reg flows. We have January to enhance it.

MariagraziaAlastra:
@jbarrett Yeah, I fear email address isn't really optional right now, I'll have to tweak a bit the endpoint.
If you're happy with it I can make further changes in other PRs and they'll get staged together with your gh login, what do you think?

jbarrett:
Yeah, we should probably make them kind of the same task since they're in the same area. I'll stage this now and we can integrate more stuff on Monday.

MariagraziaAlastra:
Cool :+1:

jagtalon:
Aw didn't get to test it :(

jagtalon:
@MariagraziaAlastra how does it work for errors?

jbarrett:
@jagtalon You can still evaluate it on staging :)

MariagraziaAlastra:
@jagtalon Checks input values on the frontend and shows/hides error messages depending on what the error is

PR #1136, opened 2015-12-18
IAPageNew.js - submit register and login forms on return keypress too

MariagraziaAlastra:
@jbarrett As requested!


MariagraziaAlastra:
@jbarrett feel free to merge if you think it's good :) I tested it too and works for me

PR #1137, opened 2015-12-20
New IA wizard - add support for multiple queries

MariagraziaAlastra:
Multiple queries can be added using commas to separate them.
E.g. "query1, query2, query3" (doesn't matter whether you put spaces between commas or not)

Also making sure the example queries are saved on IA creation.

//cc @russellholt @jagtalon


MariagraziaAlastra:
When creating the IA:
screenshot-maria duckduckgo com 5001 2015-12-20 23-38-56

After creation:
screenshot-maria duckduckgo com 5001 2015-12-20 23-39-45

PR #1142, opened 2015-12-21
Removing now-unused deploy script

jbarrett:
Just nuking this in case people are tempted to run it.

@jkv Can you confirm this script is no longer necessary? Thanks!


jkv:
@jbarrett  :+1: Not used anymore.

jbarrett:
Thanks @jkv!

PR #1156, opened 2015-12-25
Dev Pipeline - update detail sidebar info after field change

MariagraziaAlastra:
//cc @russellholt 
This makes the hide/make public button refresh after clicking on it


russellholt:
:+1:  :rocket:

Issue #904, opened 2015-08-28
Replace ->first with ->one_row in all instances?


jbarrett:
See [DBIx::Class::Helper::ResultSet::OneRow](https://metacpan.org/pod/DBIx::Class::Helper::ResultSet::OneRow) for description of possible memory leak and query inefficiencies with ->first.

PR #1158, opened 2015-12-29
Changing `first` to `one_row`

cngarrison:
Fixes: #904 

Manually changed each instance. Some queries were also setting `rows`, so removed those. 

There are also a couple of instances where the $resultset is being kept and may still need full cursor. I didn't change those; I did add some comments to explain why. From [DBIC::H::RS::OneRow](https://metacpan.org/pod/DBIx::Class::Helper::ResultSet::OneRow):

>Fundamentally the difference is that when you use one_row you are guaranteed to exhaust the underlying cursor.


jbarrett:
@cngarrison Many thanks for this! If I don't get round to reviewing this week (I'm away later) it will be first thing next week.

jbarrett:
:+1: Many thanks @cngarrison!

I'll be interested to see what effect this has on our graphs.

PR #1159, opened 2015-12-29
Support osx dev environment

cngarrison:
I had to make a couple of changes to to startup scripts to get community-platform to run in my OS X environment. The changes shouldn't cause breakages on other platforms, but I haven't tested on ubuntu (or debian). See the commit messages for details of the two changes.


jbarrett:
:+1: Thanks @cngarrison!

PR #1164, opened 2016-01-11
Wrong closing tag: Messes up the spacing

jagtalon:
Closing tag
Open all the doors and let you out into the world
Closing tag
Turn the lights up over every boy and every girl.
Closing tag
One last call for alcohol so finish your whiskey or beer.
Closing tag
You don't have to go home but you can't stay here.


MariagraziaAlastra:
Thanks for the song @jagtalon :cactus:

PR #1167, opened 2016-01-12
Discuss links now point to duckduckhack.com

jagtalon:
Instead of the forum which we don't link from anywhere anymore.

For @MariagraziaAlastra

### DIFF
--- a/templates/index.tx
+++ b/templates/index.tx
@@ -5,7 +5,7 @@
 <div class="callout">	
 	<div class="third">
 		<div class="callout-box green">
-			<a class="head" href="<: $u('Forum','index') :>">
+			<a class="head" href="http://duckduckhack.com/#join-community">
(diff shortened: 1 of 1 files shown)


templates/index.tx
```diff
@@ -5,7 +5,7 @@
 <div class="callout">	
 	<div class="third">
 		<div class="callout-box green">
-			<a class="head" href="<: $u('Forum','index') :>">
+			<a class="head" href="http://duckduckhack.com/#join-community">
```
MariagraziaAlastra: This doesn't go to the "join a community" section, though

MariagraziaAlastra: Waiting for DDH release so I can verify the section id is correct

MariagraziaAlastra:
Thanks @jagtalon, LGTM :+1:

PR #1168, opened 2016-01-13
Add author to blog feed, add link rel to feed on blog index

jbarrett:


tagawa:
Looks good except that `<link>` elements should be in the head of an HTML document. Is that possible?
Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#Notes

tagawa:
And the `rel` attribute should be `rel="alternate index"`. The `index` value tells the UA that the linked feed is an alternative version of the blog index rather than an individual post.

jbarrett:
Thanks @tagawa!

tagawa:
:+1: @john. Looks good to me!

tagawa:
And the user who reported it says:
"confirmed – awesome – thanks."

PR #1171, opened 2016-01-13
New IA Wizard - automatically set perl module and tab name for cheatsheets

MariagraziaAlastra:
//cc @jagtalon @jdorweiler 
screenshot-maria duckduckgo com 5001 2016-01-13 20-48-49


jagtalon:
:heart_eyes_cat:

Issue #1175, opened 2016-01-14
Change menu position of Community Leaders Spanish translation


tagawa:
In the left-hand menu of the help pages the [Community Leaders (es)](https://duck.co/help/community/community-leaders-es) item is far above it's English equivalent ([Community Leaders](https://duck.co/help/community/community-leaders)). Ideally it should be moved one below. The section index page is fine by the way.

jbarrett:
Thanks @tagawa!

tagawa:
Gracias to you for the new sort thingy.

PR #1178, opened 2016-01-15
Sort help articles by sort value in side menu

jbarrett:
#1175 Cc: @tagawa


tagawa:
You're on a roll this week! I think this looks good. It orders by the existing sort value from what I can tell which sounds perfect.

jbarrett:
Yep, it will be staged later today.

PR #1179, opened 2016-01-15
IA Index - add sorting via URL params

MariagraziaAlastra:
//cc @russellholt 
This allows to sort the index using either sort_asc (for sorting in ascending order) or sort_desc (descending) for any field we retrieve, including live_date and created_date.
Eg if you want to sort in a descending order by live_date: /ia?sort_desc=live_date
screenshot-maria duckduckgo com 5001 2016-01-15 14-50-01

If you want to sort as done above and filter by repo, showing only spices:
/ia?repo=spice&sort_desc=live_date
screenshot-maria duckduckgo com 5001 2016-01-15 15-34-00


russellholt:
This sorting is good but it doesn't actually solve the problem of finding new IAs released in a time period.
- can't see the dates
- can't verify that these are actually new and not updated

PR #1185, opened 2016-01-20
New IA Wizard - better error message when IA Page already exists

MariagraziaAlastra:
When creating an IA Page from a PR link:
screenshot-maria duckduckgo com 5001 2016-01-20 20-04-03

When manually filling in the fields:
screenshot-maria duckduckgo com 5001 2016-01-20 20-26-20


abeyang:
@MariagraziaAlastra can we use the red messaging for the modal as well? /cc @jagtalon

MariagraziaAlastra:
@abeyang Sure, will make a new PR!

PR #1187, opened 2016-01-21
New IA Wizard - login via GitHub

MariagraziaAlastra:
Just need to store the link in the session when the IA Page is created from a PR and I'm done here.
screenshot-maria duckduckgo com 5001 2016-01-21 22-25-55

After login and creation:
screenshot-maria duckduckgo com 5001 2016-01-21 22-28-28

### DIFF
--- a/templates/instantanswer/new_ia.tx
+++ b/templates/instantanswer/new_ia.tx
@@ -163,3 +163,6 @@
         </ul>       
     </div>
 </div>
+<: if $c.session.ia_data { :>
+    <div class="hide" id="session_ia_data"></div>
+<: } :>
--- a/templates/instantanswer/new_ia.tx
+++ b/templates/instantanswer/new_ia.tx
@@ -24,7 +24,7 @@
         <span class="clearfix"></span>
     </div>
     
-    <div id="signup-bg" class="modal-bg hide"></div>
+   <!---- <div id="signup-bg" class="modal-bg hide"></div>
     <div id="signup-form" class="hide modal-wrapper">
         <div>
             <h3 class="label left"> Create an Account</h3>
@@ -48,7 +48,7 @@
         <p class="center tx-clr--lt">Already have an account? <a class="login_newia" href="#">Sign In</a></p>
 
         <span class="clearfix"></span>
-    </div>
+    </div> --->
     
     <div id="login-bg" class="modal-bg hide"></div>
     <div id="login-form" class="hide modal-wrapper">
@@ -66,6 +66,9 @@
             <input id="login-pwd-input" type="password" placeholder="Password" name="password" class="pwd-input frm__input" />
 
         <div class="button btn--primary"  name="login" id="login-save" >Sign in</div>
+        <div>
+            <p> Or <a id="save_before_oauth" href="#">Login from GitHub</a></p>
+        </div>
         </div>
         <p>Your duck.co community account is not in any way connected to your searches on duckduckgo.com</p>
         <span class="clearfix"></span>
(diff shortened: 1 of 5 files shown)


MariagraziaAlastra:
Creation from PR link updated as well.

templates/instantanswer/new_ia.tx
```diff
@@ -154,3 +135,6 @@
         </ul>       
     </div>
 </div>
+<: if $c.session.ia_data { :>
+    <div class="hide" id="session_ia_data"></div>
+<: } :>
```
russellholt: do you prefer to use the inline syntax everywhere instead of the xslate line-oriented syntax?

```
: if $c.session.ia_data {
     <div class="hide" id="session_ia_data"></div>
: }
```

templates/instantanswer/new_ia.tx
```diff
@@ -76,25 +53,29 @@
...
+        placeholder="e.g., AlternativeTo lists alternatives to web-based software, desktop computer software, and mobile apps, and sorts the alternatives by various criteria, including the number of registered users who have clicked the Like button for each of them on AlternativeTo."><: if $c.session.ia_data { :><: $c.session.ia_data.description :> <: } :></textarea>
+        </div>
 	 
 	    <div class="wizard_field">
 	      <div class="ia-new__title tx-clr--dk">Example Query</div>
 		<p class="field_desc tx-clr--lt">List out some example queries that will trigger this Instant Answer.</p>
-		<input class="frm__input wizard_field_insert" id="example_query-input" placeholder="e.g., Alternative To Notepad, Alternative To Spotify for Windows" />
+		<input class="frm__input wizard_field_insert" id="example_query-input"  
+        <: if $c.session.ia_data { :>value="<: $c.session.ia_data.example_query :><: if $c.session.ia_data.other_queries { :>,<: for $c.session.ia_data.other_queries -> $query { :> 
+        <:$query :>, <: } :><: } :>"<: } :>
```
russellholt: pretty sure you can combine sequential xslate tags. `<: } :><: } :>` becomes `<: } } :>` if it must be inline

russellholt:
:+1:

templates/instantanswer/new_ia.tx
MariagraziaAlastra: Yup, done, thanks!

templates/instantanswer/new_ia.tx
MariagraziaAlastra: I usually go for the inline syntax if I'm inside an element (eg if I have to decide which class to add), I think it's clearer.
In this case I have to decide whether to render an element or not, and since I'm outside of it I prefer this other style.

russellholt:
I think it's better phrased as "login with github"

MariagraziaAlastra:
@russellholt Done!

russellholt:
LGTM

russellholt:
has this travis error

```
[14:52:07] t/blog.t ............... Global symbol "@feed_properites" requires explicit package name at /home/travis/perl5/perlbrew/perls/5.16/lib/site_perl/5.16.3/Dancer2/Plugin/Feed.pm line 80.
```

been a problem for a while?

jbarrett:
@russellholt  Could be to do with new Dancer2 plugin architecture - will investigate.

jbarrett:
@russellholt This tests fine with our installed stack, which isn't reflected properly on travis. We should probably nuke our package cache and reinstall from the pinto mirror.

russellholt:
Ok @jbarrett . all yours.

jbarrett:
:+1:

PR #1198, opened 2016-01-28
IA Index - better query matching

MariagraziaAlastra:
screenshot-maria duckduckgo com 5001 2016-01-28 21-50-51


MariagraziaAlastra:
//cc @moollaza

PR #1208, opened 2016-02-02
Add loading animation to the main DDH pages

jagtalon:
To reduce the perception that it's loading slowly, Chris and Abe and I thought we should add animations:

https://v.usetapes.com/HrYDIqB086

It reduces the perceived slowness of the page. 

Also:
- Changed border to box-shadow in the sidebars
- Increased opacity in the nav bar


MariagraziaAlastra:
Awesome, @jagtalon! :+1:

PR #1222, opened 2016-02-09
IA Pages - last traffic date should be equivalent to the last time the data was updated

MariagraziaAlastra:
Since we're filling with zeroes the days for which we don't have any traffic data, it's important to make sure that the last day shown is the last time the data was updated.
screenshot-maria duckduckgo com 5001 2016-02-09 15-23-16


MariagraziaAlastra:
screenshot-maria duckduckgo com 5001 2016-02-09 19-22-21

PR #1227, opened 2016-02-16
Add production_state

zachthompson:
This adds production_state to the instant_answer table and makes it available in the json feed.  The valid values are:

* `offline` - will not be loaded
* `online` - will be loaded


MariagraziaAlastra:
LGTM :+1:

PR #1229, opened 2016-02-16
views/includes/head.tx: Don't set initial scale to 1.

jagtalon:
Since the header is not made to be responsive at the moment, unsetting the initial-scale would show the "zoomed out" view which prevents breakage.

### DIFF
--- a/views/includes/head.tx
+++ b/views/includes/head.tx
@@ -12,8 +12,6 @@
     <link href="/blog/rss" type="application/atom+xml" rel="alternate index" title="DuckDuckGo Blog" />
 : }
 
-<meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1" />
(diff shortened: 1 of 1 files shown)


jagtalon:
screen shot 2016-02-16 at 6 51 19 pm

russellholt:
:-1: 

I don't like it being scaled down. The blog is the sole use of mobile here.

MariagraziaAlastra:
@jagtalon It doesn't look like it changes anything for me?
So, this is how I see the blog on my instance while on mobile (I'm using an Android phone):
screenshot_2016-02-17-15-17-58

Vs the live blog:
screenshot_2016-02-17-15-23-03

MariagraziaAlastra:
Looks good on Chrome mobile :+1:

PR #1232, opened 2016-02-17
add historical contributors

jdorweiler:
I generated an sql file that will update our developer data.  It looks good to me but I'd like other people to take a look and see if anything weird jumps out.


script/ddgc_get_prev_contributors.pl
```diff
@@ -0,0 +1,159 @@
...
+    return if $dev_json =~ /$author/ig;
+
+    my $user = $d->rs('User')->find_by_github_login($author);
+    my $data;
+
+    try{
+        if($user){
+            my $ddgc_name = $user->username;
+            # Give edit permissions to the contributor
+            $ia->add_to_users($user);
```
MariagraziaAlastra: @jdorweiler please check that the user doesn't have edit permissions already or it will error.
<pre> $ia->add_to_users($user) unless $ia->users->find($user->id) </pre>

You might also want to check whether the user is an admin, since in that case you wouldn't need to give edit permissions.

Also make sure it doesn't error on this line saying $ia is unblessed (aka don't make my same mistakes :smile: )

script/ddgc_get_prev_contributors.pl
```diff
@@ -0,0 +1,159 @@
...
+    'zeroclickinfo-spice',
+    'zeroclickinfo-goodies',
+    'zeroclickinfo-longtail',
+    'zeroclickinfo-fathead'
+);
+
+my $token = "Add your token here";
+my $gh = Net::GitHub->new(access_token => $token);
+
+my @team  = $gh->org->team_members("52625");
```
MariagraziaAlastra: @jdorweiler I get an error on this line when running the script:
"Bad credentials at script/ddgc_get_prev_contributors.pl line 43."

Am I missing anything?

jdorweiler:
@MariagraziaAlastra sorry the script is more of a one time thing I included just so we have it.  The sql is what I'm worried about.

MariagraziaAlastra:
Alright :+1:

PR #1235, opened 2016-02-18
ghIssues.pl - fix Cheat Sheets tab name and edit permissions

MariagraziaAlastra:
ghIssues used to assign the "Answer" tab to all goodies, but for Cheat Sheets it should be "Cheat Sheet" instead.
Also fixed giving edit permissions to developers.

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -176,7 +176,10 @@ sub getIssues{
                 my $is_new_ia;
                 for my $tag (@{$data->{tags}}){
                     $is_new_ia = 1 if $tag->{name} eq 'New Instant Answer';
-                    $pm = "DDG::Goodie::CheatSheets" if $tag->{name} eq 'CheatSheet';
+                     if ($tag->{name} eq 'CheatSheet') {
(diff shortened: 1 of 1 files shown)


script/ghIssues.pl
```diff
@@ -561,8 +564,11 @@ sub add_developer {
     try{
         if($user){
             my $ddgc_name = $user->username;
+            
             # Give edit permissions to the contributor
-            $ia->add_to_users($user);
+            my $ia = $d->rs('InstantAnswer')->find({id => $ia_hash->{id}});
+            $ia->add_to_users($user) unless ($ia->users->find($user->id) || $user->admin);
```
jdorweiler: Can you add a log to the catch block too?  I forgot to do that.

MariagraziaAlastra: @jdorweiler Done, please let me know if it's ok :)

jdorweiler: Cool thanks,  Can you check that `$ia` exists here before using it?  Now I'm just trying to be cautious.

MariagraziaAlastra:
@jdorweiler it should be safer now :smile:

PR #1236, opened 2016-02-18
Remove fatheads/longtails from blockgroup data

zachthompson:
Fatheads and Longtails don't need to be loaded into blocks:

1. Set all fathead/longtail blockgroups to null in instant_answer
2. Remove the option to select those blockgroups

@jdorweiler


MariagraziaAlastra:
Thanks @zachthompson! :+1:

PR #1239, opened 2016-02-23
IAPage.js - check dev milestone before attempting to show traffic data

MariagraziaAlastra:
Deprecated IAs can have traffic data, so we need to check dev milestone as well.

### DIFF
--- a/src/js/ia/IAPage.js
+++ b/src/js/ia/IAPage.js
@@ -1785,7 +1785,7 @@
                     }
                 }
 
-                if (ia_data.live.hasOwnProperty("traffic") && ia_data.live.traffic.dates.length) {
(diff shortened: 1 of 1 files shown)


src/js/ia/IAPage.js
```diff
@@ -1785,7 +1785,7 @@
                     }
                 }
 
-                if (ia_data.live.hasOwnProperty("traffic") && ia_data.live.traffic.dates.length) {
+                if ((ia_data.live.dev_milestone === "live") && ia_data.live.hasOwnProperty("traffic") && ia_data.live.traffic.dates.length) {
```
russellholt: does this include beta?

MariagraziaAlastra: What?

PR #1244, opened 2016-02-26
Redirect all requests for /wear to ddg about page

jbarrett:
Short circuiting all requests to the REDIRECT maker. Using this approach so we can quickly back out the change, should we choose to. Could also do this in nginx, but this is probably easier to back out.


MariagraziaAlastra:
:+1:

PR #1245, opened 2016-02-26
IA Pages - remove asana integration

MariagraziaAlastra:
As discussed yesterday //cc @jbarrett

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -225,15 +225,6 @@ sub dev_pipeline_json :Chained('dev_pipeline_base') :PathPart('json') :Args(0) {
 
     $key = 'dev_milestone';
 
-    my $asana_server = "http://beta.duckduckgo.com/install?asana&ia=everything";
-
-    my $result = asana_req('', $asana_server);
-    try{
-        $result = $result->decoded_content ? from_json($result->decoded_content) : undef;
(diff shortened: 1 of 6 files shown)


lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -911,79 +889,6 @@ sub send_to_beta :Chained('base') :PathPart('send_to_beta') :Args(0) {
     return $c->forward($c->view('JSON'));
 }
 
-sub asana :Chained('base') :PathPart('asana') :Args(0) {
```
jbarrett: Do we want to leave these subs in place? Or will they be moved out to a script or something at some point?

Wondering if they'll be used again in this controller.

jbarrett:
Besides comment, :+1: 

Thanks!

lib/DDGC/Web/Controller/InstantAnswer.pm
MariagraziaAlastra: It sounds like we don't really know what we want the asana integration behaviour to be, so the code might change as well in the future, even if it remains in the controller. Not sure

jbarrett: Good enough - it's in git history if we really need it.

Happy to merge?

MariagraziaAlastra: Yup thanks!

PR #1246, opened 2016-02-26
InstantAnswer.pm - only check edit permissions for traffic graph

MariagraziaAlastra:
If a user has edit permissions for an IA Page it means he's a contributor.
There's no need to check if he's actually listed as such.


jbarrett:
:+1:

PR #1249, opened 2016-02-27
Contributor Database for Engagement Testing

alohaas:
### Overview: 
This PR is a first pass at consolidating engagement-test-relevant (sorry for that elaborate compound word :v:) data from GitHub commits, pull requests, issues, and comments into the `github_event` table.

### Notes:
- I've added the `update_github_event_from_data` subroutine to the `GitHub.pm` module in DDGC. It selectively inserts data from it's parent sub into the `github_event` table.
- GitHub commits have SHAs as unique identifiers, which are alpha-numeric strings; other GitHub events have unique integer ids. Because the event's user and date are the most important for our tests now, I've chosen to consolidate the unique identifies as `text` values for the `github_id` key, accommodating both SHAs and numeric ids.

### Issues, Considerations, and Next Steps:
- This implementation does not account for events prior to the last repo events already in the database. 
- I'm recording one date for each event now – the creation date. Should non-creation dates be recorded ( PR merges, issue assignments, and issue closures) as individual events?
- Please excuse my perl etc. :grin: 

cc: @russellholt @zachthompson @kablamo @zekiel

### DIFF
--- a/lib/DDGC/DB/Result/GitHub/User.pm
+++ b/lib/DDGC/DB/Result/GitHub/User.pm
@@ -63,5 +63,9 @@ has_many github_issue_events => 'DDGC::DB::Result::GitHub::Issue::Event',
     { 'foreign.github_user_id' => 'self.id' },
     { cascade_delete => 1 };
 
+has_many github_events => 'DDGC::DB::Result::Github::Event',
+    { 'foreign.github_user_id' => 'self.id' },
+    { cascade_delete => 1 };
+
 no Moose;
 __PACKAGE__->meta->make_immutable( inline_constructor => 0 );
--- a/lib/DDGC/DB/Result/GitHub/User.pm
+++ b/lib/DDGC/DB/Result/GitHub/User.pm
@@ -63,7 +63,7 @@ has_many github_issue_events => 'DDGC::DB::Result::GitHub::Issue::Event',
     { 'foreign.github_user_id' => 'self.id' },
     { cascade_delete => 1 };
 
-has_many github_events => 'DDGC::DB::Result::Github::Event',
+has_many github_events => 'DDGC::DB::Result::GitHub::Event',
     { 'foreign.github_user_id' => 'self.id' },
     { cascade_delete => 1 };
 
--- a/lib/DDGC/GitHub.pm
+++ b/lib/DDGC/GitHub.pm
@@ -563,7 +563,7 @@ sub update_github_event_from_data {
     $columns{github_event_date} = $data->{created_at};
     
     return $gh_repo
-           ->related_resultset('github_events')
+           ->related_resultset('github_event')
            ->update_or_create(\%columns, { key => 'github_event_id' });
 }
 
--- a/lib/DDGC/GitHub.pm
+++ b/lib/DDGC/GitHub.pm
@@ -207,7 +207,6 @@ sub wanted_repos {
         duckduckgo/zeroclickinfo-longtail
         duckduckgo/p5-app-duckpan
         duckduckgo/duckduckhack-docs
-        duckduckgo/duckduckgo
     |;
 }
 
@@ -527,10 +526,11 @@ sub update_repo_forks {
     push @gh_forks, $self->update_repo_fork_from_data($gh_repo, $_)
         for @$forks_data;
     
+    print "   events...\n"; 
     my @gh_events;
     push @gh_events, $self->update_github_event_from_data($gh_repo, $_, 'github_fork') 
        for @$forks_data;
-
+    
     return \@gh_forks;
 }
 
@@ -556,15 +556,15 @@ sub update_github_event_from_data {
     my ($self, $gh_repo, $data, $eventtype) = @_;
 
     my %columns;
-    $columns{github_id}         = $data{id};
+    $columns{github_id}         = $data->{id};
     $columns{github_user_id}    = $self->find_or_update_user($data->{owner}->{login})->id;
-    $columns{github_repo_id}    = $gh_repo{id};
+    $columns{github_repo_id}    = $gh_repo->{id};
     $columns{github_event_type} = $eventtype;
-    $columns{github_event_date} = $data{created_at};
+    $columns{github_event_date} = $data->{created_at};
     
     return $gh_repo
            ->related_resultset('github_events')
-           ->update_or_create(\%columns, { key => 'github_event__id' });
+           ->update_or_create(\%columns, { key => 'github_event_id' });
 }
 
 
--- a/lib/DDGC/GitHub.pm
+++ b/lib/DDGC/GitHub.pm
@@ -526,6 +526,10 @@ sub update_repo_forks {
     my @gh_forks;
     push @gh_forks, $self->update_repo_fork_from_data($gh_repo, $_)
         for @$forks_data;
+    
+    my @gh_events;
+    push @gh_events, $self->update_github_event_from_data($gh_repo, $_, 'github_fork') 
+       for @$forks_data;
 
     return \@gh_forks;
 }
@@ -541,10 +545,27 @@ sub update_repo_fork_from_data {
     $columns{created_at}     = parse_datetime($fork->{created_at});
     $columns{updated_at}     = parse_datetime($fork->{updated_at});
     $columns{gh_data}        = $fork;
-
+    
     return $gh_repo
         ->related_resultset('github_forks')
         ->update_or_create(\%columns, { key => 'github_fork_github_id' });
 }
+
+# populate the github_event table separately while filling in the others
+sub update_github_event_from_data {
+    my ($self, $gh_repo, $data, $eventtype) = @_;
+
+    my %columns;
+    $columns{github_id}         = $data{id};
+    $columns{github_user_id}    = $self->find_or_update_user($data->{owner}->{login})->id;
+    $columns{github_repo_id}    = $gh_repo{id};
+    $columns{github_event_type} = $eventtype;
+    $columns{github_event_date} = $data{created_at};
+    
+    return $gh_repo
+           ->related_resultset('github_events')
+           ->update_or_create(\%columns, { key => 'github_event__id' });
+}
+
 
 1;
--- a/lib/DDGC/GitHub.pm
+++ b/lib/DDGC/GitHub.pm
@@ -536,7 +536,7 @@ sub update_repo_forks {
 
 sub update_repo_fork_from_data {
     my ($self, $gh_repo, $fork) = @_;
-
+    
     my %columns;
     $columns{github_id}      = $fork->{id};
     $columns{github_user_id} = $self->find_or_update_user($fork->{owner}->{login})->id;
@@ -554,17 +554,18 @@ sub update_repo_fork_from_data {
 # populate the github_event table separately while filling in the others
 sub update_github_event_from_data {
     my ($self, $gh_repo, $data, $eventtype) = @_;
-
-    my %columns;
-    $columns{github_event_id}   = $data->{id};
-    $columns{github_user_id}    = $self->find_or_update_user($data->{owner}->{login})->id;
-    $columns{github_repo_id}    = $gh_repo->{id};
-    $columns{github_event_type} = $eventtype;
-    $columns{github_event_date} = $data->{created_at};
     
+    my %columns;
+    $columns{github_id}         = $data->{id};
+    $columns{github_user_id}    = $self->find_or_update_user($data->{owner}->{login})->id;
+    $columns{github_repo_id}    = $gh_repo->id;
+    $columns{github_event_type} = $eventtype;
+    $columns{github_event_date} = parse_datetime($data->{created_at});
+
+      
     return $gh_repo
-           ->related_resultset('github_event')
-           ->update_or_create(\%columns, { key => 'github_event_id' });
+           ->related_resultset('github_events')
+           ->update_or_create(\%columns, { key => 'github_event_github_id' });
 }
 
 
--- a/lib/DDGC/GitHub.pm
+++ b/lib/DDGC/GitHub.pm
@@ -205,6 +205,9 @@ sub wanted_repos {
         duckduckgo/zeroclickinfo-fathead
         duckduckgo/zeroclickinfo-goodies
         duckduckgo/zeroclickinfo-longtail
+        duckduckgo/p5-app-duckpan
+        duckduckgo/duckduckhack-docs
+        duckduckgo/duckduckgo
     |;
 }
 
@@ -237,7 +240,7 @@ sub update_repo_from_data {
     my $gh_repo = $gh_user
         ->related_resultset('github_repos')
         ->update_or_create(\%columns, { key => 'github_repo_github_id' });
-
+    
     printf "Updating %s/%s\n", $gh_repo->owner_name, $gh_repo->repo_name;
     print "   commits...\n";
     $self->update_repo_commits($gh_repo);
--- a/lib/DDGC/GitHub.pm
+++ b/lib/DDGC/GitHub.pm
@@ -556,7 +556,7 @@ sub update_github_event_from_data {
     my ($self, $gh_repo, $data, $eventtype) = @_;
 
     my %columns;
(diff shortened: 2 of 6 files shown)


alohaas:
Hey @kablamo, since you're the original author of the primary script and similar tables in this PR, I think @jbarrett and @zachthompson think you would be best suited to review this PR. I agree :)

I am working on additional testing scripts/tables, but I will make a separate PR for those. The data in the `gihub_events` table that this PR creates would be helpful in our User Pages project so I would like to get this merged soon.

Let me know if you can help me with this! :v:

lib/DDGC/GitHub.pm
```diff
@@ -289,6 +291,10 @@ sub update_repo_pulls {
     push @gh_pulls, $self->update_repo_pull_from_data($gh_repo, $_)
         for @$pulls_data;
 
+    my @gh_events;
+    push @gh_events, $self->update_repo_pull_from_data($gh_repo, $_, 'github_pull')
```
kablamo: I think you meant `update_github_event_from_data()` here.

alohaas: You're right! In my effort to wrap this up quickly, I wasn't careful with my copy/paste.
Thanks :+1:

alohaas:
@kablamo brought up some great suggestions for ways to improve this PR and to resolve the issue of missing past events.

1. Rather than use the naming convention `github_...`, which is used for raw info from the GithHub API it makes more sense to use a different naming convention `contributor_...` or in this case `contributor_activity`. I'll need to make a new result set for this, the extra work is worth the effort since we will want `contributor_tests` soon.
2. After building and populating a table for the first time on release, I can use an sql script to backfill past events from the `github_...` tables. 
3. We currently don't have a `github_issue_comments` table. This is activity we want, so a new API query may be necessary. 
4. The `github_event_type` column could benefit from more specific strings, taking into account the possible future need to expand the types of activity we are collecting.

Thanks @kablamo as well as @jbarrett and @zachthompson for the support. More commits on the way...

zachthompson:
For past events, is there any reason we can't truncate the tables and reload everything so we get them without having to write a separate script?

alohaas:
@zachthompson i think @kablamo might be able to answer that.

sql/1.107/contributor_activity.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+CREATE TABLE contributor_activity (
+    id serial NOT NULL PRIMARY KEY,
+    github_id text NOT NULL UNIQUE,
```
zachthompson: What is the difference between `github_id` and `contributor_id`?

We probably don't want to delete events if the user or repo goes away.  Events still happened that we may want to evaluate.

alohaas: @zachthompson i was also struggling with distinguishing the two via their column name. 
`github_id` refers to the github-assigned event id or sha number
`contributor_id` is a foreign key pointing to the user who was responsible for the activity

:100: agree on the cascade issue. will change.

zachthompson: Ah ok.  I imagine it's unique and could be used for the primary key, no?

alohaas: `github_id` - yes it should be unique. does it matter that it could be an number string or a SHA mixed character string for commits (ex: 7638417db6d59f3c431d3e1f261cc637155684cd)

zachthompson: I don't think so.  The `id` is probably just extra data we don't need.

alohaas: that's a good point! So then I'll remove `Id` and set `github_id` to the Primary Key?

zachthompson: yeah, I would give that a shot and see what happens over the initial load.

alohaas: @zachthompson alrighty. I'll see how it nets out.

zachthompson: Could we make `github_id` more descriptive, e.g. `github_event_id` or something?

alohaas: I like that idea :+1:

alohaas:
@kablamo I've refactored the files to conform to the new naming convention which address the first item on the list above. 

I'm moving on to writing a backfill script, but your review of the initial changes would be appreciated :)

zachthompson:
I would see if we can't just remove the data and run the script on empty tables first.  Is this data being used live or just to do period reports?

alohaas:
@zachthompson I would like that, yes :) I think the `github_*` data is being used for the user pages, but I can't be 100% certain if it is or delicate the data is. @jbarrett probably knows and @kablamo may know other contingencies. 

Follow up thought: if we want to add the `github_issue_comments` table as well, we may want to add that before dropping the tables, but if dropping the tables ain't not thang (ideal scenario) then it doesn't matter.

zachthompson:
Just truncate, no need to drop.  Not sure if we can live without the data for that long though.  Have to see.

Are comments not included here as events? I'm not sure I follow with the new table.

alohaas:
Okay. Yeah I'm not sure how much flexibility is there. 

Yeah so comments are events and there are different types of comments we are pulling in. As I understand it, right now we have pr comment which are "review comments" and also "repo comments" which are comments on commits. We don't have issue comments _yet_.
More here : https://developer.github.com/v3/pulls/comments/

zachthompson:
Ah I see.  Probably want those, particularly for something like the maintainer test since that's where a lot of activity is happening.

alohaas:
:+1: aright. Then it seems like while we wait to find out whether or not we can trancate the tables, i should work on bringing issue comments into the db.

zachthompson:
Can't you just add the issue comments to the existing table with a different `contribution_type`?

zachthompson:
Were [access_token and users_id](https://github.com/duckduckgo/community-platform/blob/alohaas/contrib_data/lib/DDGC/DB/Result/GitHub/User.pm#L27-L28) recently removed from `github_user`?  I don't see them in master and they cause failures when trying to run this.

alohaas:
@zachthompson it looks like @jbarrett removed them. 
https://github.com/duckduckgo/community-platform/commit/472e56bf5b1b8f9465184c868575f4d990b602b9

alohaas:
argh @zachthompson i just realized we are already getting issue comments...i think. check out the gh_api query builder module that GitHub.pm uses https://github.com/duckduckgo/community-platform/blob/886b09013a03c99ebf3cb94e1426ceb731a6f550/lib/DDGC/GitHub/Cmds.pm#L60

same as https://developer.github.com/v3/issues/comments/#list-comments-in-a-repository

zachthompson:
Ok? Less work for you I guess.  Are they being recorded in your new table yet?

alohaas:
They're actually in the `repo_comments` table already...very confusing.

lib/DDGC/DB/Result/GitHub/User.pm
```diff
@@ -24,6 +24,8 @@ column updated_at             => {data_type => 'timestamp with time zone', is_nu
 column isa_owners_team_member => {data_type => 'int',    is_nullable       => 0, default_value => 0};
 column scope_public_repo      => {data_type => 'int',    is_nullable       => 0, default_value => 0};
 column scope_user_email       => {data_type => 'int',    is_nullable       => 0, default_value => 0};
+column access_token           => {data_type => 'text',   is_nullable       => 1};
```
zachthompson: These are back.  Also looks like it has conflicts now.

alohaas:
@kablamo we are getting closer, but a couple of things remain:

T `contributor_activity` backfill – Earlier in the PR comments, @zachthompson questioned whether we could truncate the `github_pull`, `github_review_comment`, `github_repo_comment`, `github_repo_commit`, and `github_issue` tables and repopulate them along with `contributor_activity` rather than writing a backfill script. We aren't sure if this will break anything...

Based on some recent conversations around maintainership, we also may want to expand our data to include the `/$repo/issues/events` endpoint.

kablamo:
@alohaas that will probably work.  I don't think anyone uses data from these tables atm.  Even if they do, I don't think the generated ids (like github_pull.id) are used anywhere.  But I'm not 100% sure.  I guess make some backups before you do it.  The ids are the only thing I can think of that I might worry about.  Writing the sql backfill script should be pretty simple I would think.  But probably not as simple as truncating the table.  

How to pull from a new endpoint is not obvious.  Let me know if you get stuck.  Basically you will want to add a sub similar to this: 
https://github.com/duckduckgo/community-platform/blob/master/lib/DDGC/GitHub/Cmds.pm#L77
and of course this as well:
https://github.com/duckduckgo/community-platform/blob/master/lib/DDGC/GitHub.pm#L316

alohaas:
#### Refactor + Enhancements To-Do List

- [x] github api method for issue events
- [x] github api method for commit comments
- [ ] add schema for issue events
- [ ] add schema for commit comments
- [ ] add method to update issue event table
- [ ] add method to update commit comment table
- [ ] populate contributor_activity with issue events
- [ ] populate contributor_activity with commit comments

- [ ] **add/refactor support for new event types**
- [ ] git_commit (was github_commit)
- - -
- [ ] git_issue_assigned
- [ ] git_issue_unassigned
- [ ] git_issue_labeled
- [ ] git_issue_unlabeled
- [ ] git_issue_opened (was github_issue_update and github_pull)
- [ ] git_issue_closed
- [ ] git_issue_merged
- [ ] git_issue_reopened
- [ ] git_ issue_synchronized
- - -
- [ ] git_review_comment (was github_review_comment)
- [ ] git_issue_comment (was github_repo_comment)
- [ ] git_commit_comment
- - -
- [ ] git_pull_assigned
- [ ] git_pull_unassigned
- [ ] git_pull_labeled
- [ ] git_pull_unlabeled
- [ ] git_pull_opened (was github_issue_update and github_pull)
- [ ] git_pull_closed
- [ ] git_pull_merged
- [ ] git_pull_reopened
- [ ] git_ pull_synchronized

zachthompson:
I added a where clause to exclude repos that weren't zeroclickinfo-* and saw a pretty big performance hit.  Probably want to index it.  Same with contribution_date.  Aggregate cost ~3678 -> ~122 with both.

alohaas:
@zachthompson and @kablamo I believe I have added most of the changes which we need. clearly there are bugs, which I will be looking into, but I wanted to give a heads up that it's about ready for another review.

sql/1.107/github_commit_comments_table.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+
+CREATE TABLE "github_commit_comment" (
+    "id" serial NOT NULL,
```
zachthompson: Why id?

sql/1.107/github_commit_comments_table.sql
```diff
@@ -0,0 +1,18 @@
...
+    "github_repo_id" bigint NOT NULL REFERENCES github_repo(id) ON DELETE CASCADE,
+    "github_user_id" bigint NOT NULL REFERENCES github_user(id) ON DELETE CASCADE,
+    "position" bigint NOT NULL,
+    "line" bigint NOT NULL,
+    "number" bigint NOT NULL,
+    "body" text NOT NULL,
+    "created_at" timestamp with time zone NOT NULL,
+    "gh_data" text NOT NULL,
+    
+    PRIMARY KEY (commit_id, number)
```
zachthompson: Where is commit_id?

alohaas: ah, i changed commit_id to sha and it looks like i forgot to make the change here.

sql/1.107/github_commit_comments_table.sql
alohaas: I was following the pattern of other `github_*` table creation scripts, like this one.
https://github.com/duckduckgo/community-platform/blob/886b09013a03c99ebf3cb94e1426ceb731a6f550/sql/1.044/github_tables.sql

alohaas: @zachthompson if you think it's not necessary, i can remove it

zachthompson: Right, I get that.   As with `contributor_activity` it's just extra data since it's not being used to, say, make a primary key.  Not a big deal just wondering.

sql/1.107/github_commit_comments_table.sql
alohaas: @zachthompson there are a few more errors relating to commit comments i'm seeing like in the 
`update_repo_commit_comment_from_data` in `GitHub.pm`. These are all relating to me thinking through how to identify commit comments unique constraints. I think we should be using the `sha` of the comment's commit and the `commit_id` which would be unique to that comment within the commit, if that makes sense...I need to remove the `number` key and replace with `commit_id` in a few places. Here's the API reference if you're interested - https://developer.github.com/v3/repos/comments/#list-commit-comments-for-a-repository

zachthompson:
Is this ready to be tested?

sql/1.107/github_issue_events_table.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+    
+    ALTER TABLE github_event_issue DROP COLUMN created;
```
zachthompson: Is this supposed to be github_issue_event?

lib/DDGC/DB/Result/GitHub/CommitComment.pm
```diff
@@ -0,0 +1,32 @@
...
+    is_nullable => 0, 
+    serializer_class => 'JSON',
+    serializer_options => { convert_blessed => 1, pretty => 1 }, 
+};
+
+unique_constraint [qw( sha comment_id)];
+
+belongs_to 'commit', 'DDGC::DB::Result::GitHub::Commit', 'sha';
+belongs_to 'repo', 'DDGC::DB::Result::GitHub::Repo', 'github_repo_id';
+belongs_to 'user', 'DDGC::DB::Result::GitHub::User', 'contributor_id';
```
zachthompson: No `contributor_id` in user.

sql/1.107/github_issue_events_table.sql
alohaas: gah yes.

lib/DDGC/DB/Result/GitHub/CommitComment.pm
alohaas: you're right. the similarities between the names within these data types are killing me.

alohaas:
@zachthompson it should be ready to test and debug as those other tests aren't passing :-1:

if you're starting with a contributor activity table already...
1. run `sql/1.107/github_issue_events_table.sql` to alter the issue events table
2. run  `sql/1.107/github_issue_events_table.sql` to create the commit comments table
3. run `lib/DDGC/GitHub.pm` and hope for the best.

I'll be working through this as well.

zachthompson:
I couldn't get it to run.  Have you been able to run through to completion with the new changes?  Let me know when you do and I'll check out the data.

zachthompson:
Using these id fields on other tables as foreign keys has been problematic.  When they get rebuilt, the ids change unless the sequence is reset.

alohaas:
@zachthompson which fields would be better to use?

zachthompson:
githhub_user.github_id and github_repo.github_id.

zachthompson:
Additional observations about these fields:

1. If we fkey to github_user, we can't add activity unrelated to github.  I know @russellholt was considering updating metadata on the complat as activity in the future.
2. If the repo can't be null, we have the same issue.  We might think of it like a `where` or `source` of the activity.

alohaas:
1. @zachthompson isn't authenticating with github on the platform a move towards using a github identify as a constant for users? As an example, if we wanted to pull in slack activity, we could use the `users` table to identify the corresponding `github_user.github_id` when the value is assigned to the `contributor_activity` record, but only if the `slack_id` were a key in the `users` table. Otherwise, do you think we should use the `users` rather than the `github_user`table?
2. Should there be an additional key for the source, with the option of leaving the `github_repo` null? I wouldn't want to lose the foreign key unless that doesn't actually matter and we could populate the name of the repo instead.

zachthompson:
How are we doing here?

PR #1251, opened 2016-02-27
Migrations: The  DB can't be created unless the local user has peer/trust auth with Postgres

mintsoft:
So, previously it assumed that either the `dropdb` command wouldn't be available or the user running the script would have `dropdb`/`createdb` permissions using peer/trust authentication on the local install of postgres.

That's not the case for me, so if it fails to execute the `dropdb` (for example due to: `dropdb: could not connect to database template1: FATAL:  Peer authentication failed for user "ddgc"` then it will now flip over to the DBI connection.

Additionally, if the db didn't exist then the DBI path would fail, the drop is now defensive

### DIFF
--- a/script/ddgc_deploy_dev.pl
+++ b/script/ddgc_deploy_dev.pl
@@ -36,13 +36,13 @@ if ($kill) {
 			my $userarg = length($config->db_user) ? "-U ".$config->db_user : "";
 			print "Truncating database...\n";
 			system("dropdb $userarg ".$db);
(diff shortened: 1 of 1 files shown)


jbarrett:
Thanks @mintsoft! Will get back to you on this later today.

GuiltyDolphin:
@jbarrett Had the same issue as @mintsoft, cloned this down and it worked like a charm! :dolphin:

jbarrett:
Thanks @mintsoft and @GuiltyDolphin!

PR #1256, opened 2016-02-29
Update maintainers based on historical information

zachthompson:
Script to update the maintainer field for live instant answers based on historical data.

### DIFF
--- a/sql/1.152/maintainer_update.pl
+++ b/sql/1.152/maintainer_update.pl
@@ -14,19 +14,24 @@ my %seen;
 while(<DATA>){
 	chomp;
 	my ($id, $git_login) = split /\t/;
-	unless(exists $seen{$git_login}){
+	unless($seen{$git_login}){
 		if(my $git_user = $d->rs('GitHub::User')->find({login => $git_login})){
 			if(my $duckco_user = $d->rs('User')->find({github_id => $git_user->github_id})){
-				$seen{$git_login} = $duckco_user->username;
+				$seen{$git_login} = $duckco_user;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


MariagraziaAlastra:
What about giving them edit permissions for their IA Page, too?

zachthompson:
Is that some flag we need to set in DB or...?  I didn't have that in the scope of this task.

MariagraziaAlastra:
<pre>$ia->add_to_users($user)</pre>
Like it's done [here](https://github.com/duckduckgo/community-platform/blob/master/script/ghIssues.pl#L579), but I can add that, nbd

zachthompson:
Ok, let me check it out. I can do that.

zachthompson:
@MariagraziaAlastra  Added `add_to_user`.  Seems to work.

zachthompson:
Got it, thanks!

sql/1.152/maintainer_update.pl
```diff
@@ -0,0 +1,768 @@
...
+			$maintainer = qq|{"github":"$git_login"}|;
+		}
+
+		$ia->update({maintainer => $maintainer});
+	}
+	else{ warn "Failed to lookup IA with id of $id" }
+}
+
+__DATA__
+us_government_agencies_cheat_sheet	aasoliz
```
MariagraziaAlastra: These are all github usernames, right?

zachthompson: Yep.

MariagraziaAlastra:
Tested and LGTM, thanks! :+1:

PR #1266, opened 2016-03-08
Only add github user to maintainer

zachthompson:
Remove addition of duck.co user to maintainer field.


MariagraziaAlastra:
This needs to be removed in the backend code too, or it will still be possible to have duck.co-only maintainers. I'll do that though!

MariagraziaAlastra:
LGTM :+1:

PR #1267, opened 2016-03-08
Porting DDH::UserPage to community-platform

jbarrett:
This is almost the same as the original project - `bin/generate.pl` creates pages. `bin/dev_server.pl` is a static file server.

Differences:

- Gather has a db handle via a `DDGC`instance (so we can probably lose the HTTP operations).
- dev_server mounts static files (CSS etc)
- userpage/index is rendered within a duck.co style frame with nav bar and no account info.

Cc: @abeyang @jagtalon

### DIFF
--- a/bin/dev_server.psgi
+++ b/bin/dev_server.psgi
@@ -6,8 +6,10 @@ use warnings;
 use FindBin;
 use Plack::Builder;
 use Plack::App::Directory;
+use Plack::App::File;
 
 builder {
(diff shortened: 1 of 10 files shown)


MariagraziaAlastra:
LGTM, I just pushed my changes to your branch!
There's clearly a lot of room for improvement, but it does what it has to for now :)

jbarrett:
Thanks @MariagraziaAlastra

PR #1268, opened 2016-03-08
maintainer bug 

zachthompson:
Bug in logic, git should be added for all.
Removed the OS IAs maintained by duckduckgo; they should be unmaintained.


MariagraziaAlastra:
:+1:

PR #1272, opened 2016-03-10
Userpages - dump JSON data for each user

MariagraziaAlastra:
This change makes it possible to see the raw JSON data for each user, using the path /username/json
(eg '/jbarrett/json')

### DIFF
--- a/lib/DDH/UserPage/Generate.pm
+++ b/lib/DDH/UserPage/Generate.pm
@@ -113,15 +113,8 @@ sub generate {
 
         # JSON dump of user's data
         my $json_build_dir = $self->json_dir( $contributor );
-        open my $jfh, '>:encoding(UTF-8)', "$json_build_dir/index.html" or die();
-        print $jfh $self->xslate->render(
-            'userpage/json_index.tx',
-            {
-                json_data => $self->json->encode(
-                    $contributor_data
-                ),
-            }
-        );
+        open my $jfh, '>:encoding(UTF-8)', "$json_build_dir/index.json" or die();
+        print $jfh $self->json->encode( $contributor_data );
         close $jfh;
     }
 }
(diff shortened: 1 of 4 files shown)


lib/DDH/UserPage/Generate.pm
```diff
@@ -105,6 +110,19 @@ sub generate {
             }
         );
         close $fh;
+
+        # JSON dump of user's data
+        my $json_build_dir = $self->json_dir( $contributor );
+        open my $jfh, '>:encoding(UTF-8)', "$json_build_dir/index.html" or die();
+        print $jfh $self->xslate->render(
```
jbarrett: Wondering if we could just dump index.json alongside index.html without a template? i.e. just the raw JSON as we get it?

MariagraziaAlastra: @jbarrett Done, but for some reason it doesn't get nicely formatted like when I was using a template :(

lib/DDH/UserPage/Generate.pm
```diff
@@ -105,6 +110,12 @@ sub generate {
             }
         );
         close $fh;
+
+        # JSON dump of user's data
+        my $json_build_dir = $self->json_dir( $contributor );
+        open my $jfh, '>:encoding(UTF-8)', "$json_build_dir/index.json" or die();
```
jbarrett: Reckon it would be OK to just use `$build_dir`

jbarrett: Reckon it would be OK to just use `$build_dir`?

MariagraziaAlastra: Yeah, seems like it is, thanks!

jbarrett:
:+1: Thanks, @MariagraziaAlastra!

PR #1276, opened 2016-03-14
Generate pages when views/ dir changes

jbarrett:
@jagtalon Could you please confirm this does what you need?


jbarrett:
This re-loads the index from the generated JSON so it will keep feeding the same data back into itself. Run generate.pl to get new data (or make no changes in views/ for 15 minutes).

This should cut down on HTTP requests and DB hits and keep things relatively speedy.

jagtalon:
Perfect, thanks! :+1:

PR #1281, opened 2016-03-17
Userpages - don't include js.tx (fixes broken link redirects)

MariagraziaAlastra:
Now the links redirect to the proper pages :)


jbarrett:
:+1: Thanks @MariagraziaAlastra!

Issue #1257, opened 2016-03-03
Weird errors with ddgc_populate_ia_dev


GuiltyDolphin:
@jbarrett I was trying to run the `ddgc_populate_ia_dev.pl` script to get up-to-date with duck.co (as suggested by @jdorweiler), but I kept running into this issue:

![screenshot from 2016-03-03 17 25 27](https://cloud.githubusercontent.com/assets/8598426/13502933/ef343c68-e164-11e5-9fdd-cc76263213e1.png)

It doesn't appear to be dependency related - I've updated fairly recently, and the script is definitely getting part-way through (I'm seeing a few IAs pop up locally - as well as having put some printing in there to confirm) - but it seems to then just give up!

Any ideas?

jdorweiler:
I'm seeing a similar error when I run it.
````
Use of uninitialized value in concatenation (.) or string at /home/ddgc/community-platform/script/../lib/DDGC/DB/Base/Result.pm line 280.
SQL::Abstract::puke(): [SQL::Abstract::__ANON__] Fatal: Operator calls in update must be in the form { -op => $arg } at /home/ddgc/community-platform/script/../lib/DDGC/DB/Result/InstantAnswer.pm line 549
DBIx::Class::Storage::TxnScopeGuard::DESTROY(): A DBIx::Class::Storage::TxnScopeGuard went out of scope without explicit commit or error. Rolling back. at /opt/perlbrew/perls/perl-5.16.3/lib/site_perl/5.16.3/DBIx/Class/Exception.pm line 77
````

**Comment by GuiltyDolphin** [comment]
@jbarrett @jdorweiler I think it is to do with the maintainer field.

I checked the `$data->{$ia_id}` hash, and for the ones that were getting through the `maintainer` field was undefined, but it failed whenever one came up in the form:

```Perl
maintainer => {
    duckco => ...,
    github => ...
}
```

I set the `maintainer` field to `undef` for each of them, and it ran *much* more smoothly. I guess that explains why it was complaining about a hash!

MariagraziaAlastra:
Fixed on #1284. Thanks @GuiltyDolphin for the bug report!

GuiltyDolphin:
@MariagraziaAlastra @jbarrett Just tried it (#1284) out locally and I'm getting a different error now:

[...]

PR #1284, opened 2016-03-17
Fix ddgc_populate_ia_dev

jbarrett:
Pulling maintainer hash from data before insert.

Still barfs a load of "uninitialized value in concatenation" errors from this [stringification override](https://github.com/duckduckgo/community-platform/blob/9d0f94d5ace1219b87ae784ef00bc603a9859395/lib/DDGC/DB/Base/Result.pm#L278-L281) but I'm not too worried about that for now.

### DIFF
--- a/script/ddgc_populate_ia_dev.pl
+++ b/script/ddgc_populate_ia_dev.pl
@@ -25,6 +25,7 @@ for my $ia_id (keys $data) {
     my $topics = delete $data->{$ia_id}->{topic};
     my $developer = delete $data->{$ia_id}->{developer};
     my $attribution = delete $data->{$ia_id}->{attribution};
+    my $maintainer = delete $data->{$ia_id}->{maintainer};
     my $src_options = delete $data->{$ia_id}->{src_options};
 
     my $ia = $d->rs('InstantAnswer')->update_or_create({
@@ -38,6 +39,9 @@ for my $ia_id (keys $data) {
             : (),
         ( $developer )
             ? ( developer => encode_json( $developer ) )
+            : (),
+        ( $maintainer )
+            ? ( maintainer => encode_json( $maintainer ) )
             : (),
     });
 


MariagraziaAlastra:
Seems like it actually updated stuff, although I have an IA Page for a Cheat Sheet for which I changed the repo to "Spice" previously, and althought I do have the correct IA Page for that same Cheat Sheet, the old one with the faulty data wasn't removed. I suppose this is because the meta_id changed:
screenshot-maria duckduckgo com 5001 2016-03-17 19-41-53

Not sure how to avoid duplicates without deleting all the InstantAnswer rows before populating the db again (and I'm not a fan of this strategy). It works well overall, so I'd be happy to merge!

jbarrett:
Thanks @MariagraziaAlastra - since meta_id isn't in the repo JSON I ended up [just using id](https://github.com/duckduckgo/community-platform/blob/bf3ac0f/script/ddgc_populate_ia_dev.pl#L32)

MariagraziaAlastra:
@jbarrett [Actually...](https://github.com/duckduckgo/community-platform/blob/master/lib/DDGC/DB/Result/InstantAnswer.pm#L33)
So the id in the repo JSON is the meta_id. The id is never shown on JSON endpoints.
See also [here](https://github.com/duckduckgo/community-platform/blob/master/lib/DDGC/Web/Controller/InstantAnswer.pm#L112)

GuiltyDolphin:
@jbarrett This for #1257? If so, *sweet!*

jbarrett:
@GuiltyDolphin Don't know how I absolutely missed this report, but yes this should resolve #1257.

You guys did all the leg work, then I did it again!

MariagraziaAlastra:
I think it's good to go. Thanks @jbarrett!

GuiltyDolphin:
@jbarrett Woo! I'll check it out.

PR #1288, opened 2016-03-21
IA Pages - add 'Maintain this IA' button

MariagraziaAlastra:
Live IA without maintainer (I previously removed it):
screenshot-maria duckduckgo com 5001 2016-03-21 21-25-54

After clicking on the button:
screenshot-maria duckduckgo com 5001 2016-03-21 21-28-34

TODO: open up editing to all logged in users, not just ones with edit permissions

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -949,6 +949,7 @@ sub save_edit :Chained('base') :PathPart('save') :Args(0) {
     my $is_admin = 0;
     my $saved = 0;
     my $field = $c->req->params->{field};
+            my $value = $c->req->params->{value};
     my $msg = "";
 
     # if the update fails because of invalid values
@@ -970,11 +971,11 @@ sub save_edit :Chained('base') :PathPart('save') :Args(0) {
        $permissions = $ia->users->find($c->user->id);
        $is_admin = $c->user->admin;
 
-        if ($permissions || $is_admin) {
+        # Allow any logged in user to self-assign an IA for maintainership
+        if ($permissions || $is_admin || (($field eq 'maintainer') && ($value eq $c->user->username) && (!$ia_data->{maintainer}->{github}))) {
             $result->{is_admin} = $is_admin;
             $c->stash->{x}->{result} = $result;
 
-            my $value = $c->req->params->{value};
             my $autocommit = $c->req->params->{autocommit};
             my $rs_user = $c->d->rs('User');
             my $complat_user = $rs_user->find({username => $value}) || $rs_user->find_by_github_login($value);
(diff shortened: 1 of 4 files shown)


zachthompson:
Cool. I would think just replace "No maintainer yet" with the button.

MariagraziaAlastra:
Done:
screenshot-maria duckduckgo com 5001 2016-03-22 21-13-32

MariagraziaAlastra:
@zachthompson Ready for review

zachthompson:
If you don't have a github account it adds the duck.co account.  However, when you look at it in edit the maintainer is blank.  Probably shouldn't allow the maintainer to be set without a gh account.

MariagraziaAlastra:
@zachthompson 
screenshot-maria duckduckgo com 5001 2016-03-23 18-55-12

Right, duck.co users aren't allowed, it's just it was mistaking TestOne's username for a GitHub account (since that username is on GH for reals). I fixed the logic, let me know if you encounter any more issues :smile:

lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -1032,7 +1033,16 @@ sub save_edit :Chained('base') :PathPart('save') :Args(0) {
             if ($field =~ /designer|producer/){
                 return $c->forward($c->view('JSON')) unless ($complat_user_admin || $value eq '');
             } elsif ($field eq "maintainer") {
-                return $c->forward($c->view('JSON')) unless $value = format_maintainer($c, $value, $ia, 0);
+                my $result = format_maintainer($c, $value, $ia, 0);
+               use Data::Dumper;
+               print Dumper $result;
```
zachthompson: Maria, was just looking at this again.  Lines 1037-8 probably need to be removed.

On line 1039, what exactly are you trying to catch here?  A failed return from `format_maintainer`?  I think the logic here might be off:
``` Perl
$VAR1 = {
          'msg' => 'User has no GitHub account linked',
          'maintainer' => {}
        };
```
`maintainer` is always there, even if an empty hash, so it never gets in this `if`.  Is this why, when there is no github account, I get no error?  I see the message from `format_maintainer` come through but it never displays on the page afaict.

It seems to work fine if a user has a github login associated with their account.

zachthompson: The error message only shows for admin accounts.

MariagraziaAlastra: Yo sorry, pushed off to do other stuff, I'm looking at this again too now!

MariagraziaAlastra: Yeah you're right, it will never pass that if! I think I might have made some change at some point to make the maintainer undefined when there's no github account and then undid it, and forgot about the implications...

Fixing it!

MariagraziaAlastra:
@zachthompson Ready again for review! With @jbarrett we thought it could be nice to show a "Log in with GitHub" message for users who don't have a GitHub account linked, instead of the button, since they wouldn't be eligible as maintainers anyway. What do you think?
screenshot-maria duckduckgo com 5001 2016-06-16 16-37-26
screenshot-maria duckduckgo com 5001 2016-06-16 16-39-13
screenshot-maria duckduckgo com 5001 2016-06-16 16-42-04

MariagraziaAlastra:
Closing as not relevant anymore.

PR #1289, opened 2016-03-22
IA Pages - restrict changing dev_milestone to planning, ghosted and deprecated when production_state is online

MariagraziaAlastra:
screenshot-maria duckduckgo com 5001 2016-03-22 16-18-38


MariagraziaAlastra:
On dev pages, though, there's still no visual clue about whether the dev_milestone was updated or not. It does effectively restric those values for online IAs, but from a UX point of view it's confusing and seems buggy.
I have to find a solution for that...

MariagraziaAlastra:
@zachthompson Now the Dev Pages should revert to the current dev_milestone if the change wasn't saved. Still no error message there, but at least it should be clear the new value wasn't saved.

zachthompson:
Works for me.

PR #1290, opened 2016-03-22
InstantAnswer.pm - sort traffic data by date

MariagraziaAlastra:
On the frontend, when looping over the data, we add zeroes whenever we see date gaps to fill.
It seems like we don't always get the data sorted by date though, so we need to user order_by when querying the DB to make sure we don't add zeroes where there's no need for them.


zachthompson:
Thanks.

PR #1291, opened 2016-03-23
Move build dir to persistent directory

jbarrett:
- moving pages to ~/ddgc - we can serve from here, deployments won't touch this.


jagtalon:
Works! :+1:

PR #1292, opened 2016-03-25
Use Core team instead of Owners team.  Github made us rename Owners team name

kablamo:
cc @zachthompson


russellholt:
cc @jdorweiler

PR #1296, opened 2016-03-30
User Pages

jagtalon:
There's a lot going on here, but here are the important files:

root/static/css/userpages/UserPage.scss
root/static/js/userpages/UserPage.js
root/static/pages/ddh_up_index.html
root/static/partials/c_instantanswer.html
root/static/partials/c_issue.html


jagtalon:
We have the necessary markup. All we need to do is to make it use the right data

screen shot 2016-03-30 at 11 04 52 am

jagtalon:
screen shot 2016-03-30 at 2 49 45 pm

jagtalon:
Taking shape:

screen shot 2016-03-31 at 11 51 54 am

root/static/pages/ddh_up_index.html
```diff
@@ -1,2 +1,125 @@
...
+<div class="container">
+
+  <!-- Sidebar -->
+  <aside>
+    <div class="profile card">
+      <img src="{{response.gh_data.avatar_url}}" />
+      <div class="content">
+	<h3>{{user.name}}</h3>
+	<div class="gray">
+	  <a href="http://github.com/{{response.gh_data.login}}">@{{response.gh_data.login}}</a>
```
GuiltyDolphin: @jagtalon Can we switch this to `https`?

views/userpage/index.tx
```diff
@@ -1,103 +1,125 @@
...
+<div class="container">
+
+  <!-- Sidebar -->
+  <aside>
+    <div class="profile card">
+      <img src="{{response.gh_data.avatar_url}}" />
+      <div class="content">
+	<h3>{{user.name}}</h3>
+	<div class="gray">
+	  <a href="http://github.com/{{response.gh_data.login}}">@{{response.gh_data.login}}</a>
```
GuiltyDolphin: @jagtalon This one too (to `https`). Is there a reason for having it as `http` in the first place?

views/userpage/index.tx
```diff
@@ -1,103 +1,125 @@
...
+	  <br />{{response.gh_data.location}}
+	  </div>
+
+	</div>
+</div>
+    <div class="topic typeC" ng-repeat="(key, value) in response.topics | orderBy:'-value'">
```
GuiltyDolphin: It'd be awesome if the topics were clickable to alter which IAs were displaying ('filter by topic').

root/static/pages/ddh_up_index.html
jagtalon: @GuiltyDolphin yup, can do!

jagtalon:
@MariagraziaAlastra @bsstoner I changed the definition of `random` because we want it to return the same value every time the function is called. AngularJS thinks that the model is changing thus triggering an update, and sees that the model has changed again which triggers an update until we reach an infinite loop.

See `infdig` error: http://is.gd/Zbiz6o

root/static/pages/ddh_up_index.html
```diff
@@ -1,2 +1,297 @@
...
+<div class="container">
+
+  <!-- Sidebar -->
+  <aside>
+    <div class="profile card">
+      <img src="{{response.gh_data.avatar_url}}" />
+      <div class="content">
+	<h3>{{response.gh_data.name}}</h3>
+	<div class="gray">
+	  <a href="http://github.com/{{response.gh_data.login}}">@{{response.gh_data.login}}</a>
```
GuiltyDolphin: @jagtalon Looks like this one is still `http`?

MariagraziaAlastra:
@jagtalon Shouldn't that function trigger only once, on page load?

root/static/pages/ddh_up_index.html
jagtalon: Fixed!

jagtalon:
@MariagraziaAlastra It's inside the controller and from what I understand gets re-rendered when it's not [in sync](https://www.youtube.com/watch?v=Eo-KmOd3i7s) with the model

MariagraziaAlastra:
@jagtalon ok, thanks :smile:

MariagraziaAlastra:
@jagtalon Maybe [one time binding](https://docs.angularjs.org/guide/expression#one-time-binding) would solve this too? I need to make a new PR anyway, so I can try that if you wish

jagtalon:
@MariagraziaAlastra one-time binding seems like it would work too. Anything that would prevent the digest from re-calculation I think is a good solution

alohaas:
@jagtalon here I go!

jagtalon:
@alohaas down the rabbit hole! 🐇

root/static/partials/c_instantanswer.html
```diff
@@ -0,0 +1,24 @@
+<!-- Outside is wrapped within <div class="ia..."> -->
+
+<div class="grid-6">
+  <strong><a href="https://duck.co/ia/view/{{ia.id}}" ng-click="addImg('iapage', ia.id)">{{ia.name}}</a></strong>
```
alohaas: @jagtalon I just tried to click through to Ben's Markdown Reference IA and it redirected me to the IA index. Maybe the page doesn't exist? maybe the url is being built incorrectly?
screen shot 2016-04-12 at 3 53 31 pm

http://jag.duckduckgo.com:5002/u/guiltydolphin ( Markdownreference is the last IA in his maintaining list)

jagtalon: @alohaas it looks like the data that I have is out of data--that Markdownreference one has been ghosted and shouldn't show up at all (it's not showing up here: https://duck.co/u/guiltydolphin/). Thanks!

alohaas:
@jagtalon I'm curious to know why you've put the github iconfont files in the `root/static/css/userpages/` directory. Should they not live within a `root/static/fonts/userpages/`directory?

jagtalon:
@alohaas oh no reason. Probably should be in the fonts directory. I could move it to fonts/

alohaas:
@jagtalon I think that would be a better home for them. I would add a `userpages` subdirecty in the `fonts` folder and put them in there to maintain consistency across directories. The more orderly and intentional we can be here, the more it will help our flexibility goals.

jagtalon:
@alohaas cool now it's in the fonts/ dir

root/static/pages/ddh_up_index.html
```diff
@@ -1,2 +1,296 @@
...
+	<div class="gray">
+	  <a href="https://github.com/{{response.gh_data.login}}">@{{response.gh_data.login}}</a>
+	  <br />{{response.gh_data.location}}
+	  </div>
+
+	</div>
+</div>
+    <div class="topic typeC" ng-repeat="topic in topics | orderBy:'-count'">
+      <span class="amt"> {{topic.count}}</span>
+      <span> {{topic.topic}}</span>
```
alohaas: Mr. Chris has a million topics, but they grow dangerously close to the footer of his page.Add some padding?

screen shot 2016-04-13 at 7 54 35 am

jagtalon: @alohaas good call. I got it

jagtalon:
@alohaas fixed!

alohaas:
@jagtalon I'm seeing a double scroll bar on firefox. When I removed `overflow-x: hidden` from the body element `body` the second scroll bar disappears. Is there a reason that style was applied? Removing it seems not to break anything.

screen shot 2016-04-13 at 10 32 37 am

alohaas:
@jagtalon the user avatars adjacent to IAs aren't sitting in the center of their circles.

screen shot 2016-04-13 at 10 37 46 am

jagtalon:
@alohaas cool should be fixed!

root/static/pages/ddh_up_index.html
alohaas: @jagtalon Chris only has 21 topics in his list, which is more than most contributors, but you could have more than that. For example the `min-height` fix isn't doing to work if someone has 23 Topics. We will want to add an increased bottom margin or padding, whichever you're using to separate the topic items. I would suggest that the distance of the last topic and the footer be equal to the distance between the avatar and the first topic in the list.

root/static/js/userpages/UserPage.js
```diff
@@ -0,0 +1,166 @@
...
+        var randomNum = Math.ceil(Math.random() * 1e7);
+        pre_id = pre_id? '_' + pre_id : '';
+        element_id = element_id? '_' + element_id.replace(/[^0-9A-Za-z]/g, '') : '';
+        var imgUrlSuffix = pre_id + element_id + '?' + randomNum;
+
+        $scope.newImg = new Image();
+        $scope.newImg.src = $scope.imgUrl + imgUrlSuffix;
+	};
+
+	$scope.ias = [];
```
alohaas: Indentation starts to get a little funky here through line 111.

root/static/js/userpages/UserPage.js
```diff
@@ -0,0 +1,166 @@
...
+	// given an array, search for an attr (like 'username'): if found, return object
+	findByAttr: function(arr, attr, name) {
+	    return _.find(arr, function(obj) {
+		return obj[attr] == name;
+	    });
+	},
+
+    // return the first char in a dev's username (for when avatar is not available)
+    firstUsernameChar: function(username) {
+        username = username? username.charAt(0) : '';
```
alohaas: nitpick but that `?` needs a space before it :grimacing:

jagtalon:
@alohaas auto-indented and fixed the space

GuiltyDolphin:
@jagtalon @alohaas When I refresh the page on https://duck.co/u/guiltydolphin/ I very briefly see a menu appear to the right, this is the best screenshot I've managed to get of it:

screenshot from 2016-04-13 18 12 54

I don't see any way to expand it - any ideas?

alohaas:
@GuiltyDolphin thanks for pointing that out – it's a vestige from an earlier version of the design. We'll be removing it.

jagtalon:
@GuiltyDolphin thanks! What @alohaas said. I'll shake it off from the code

giphy

jagtalon:
@alohaas removed the slideout, although still figuring out the min-height thing. Can I postpone that fix?

alohaas:
@jagtalon I think it's OK if we postpone the fix. Next round of larger changes might warrant a revisit.

nilnilnil:
What is the testing strategy for this? How will we validate correctness in the future?

PR #1297, opened 2016-03-30
Single index file for all users.

jbarrett:
- New script, markup.pl, symlinks all index.html files to one
- New file, users.json, created.
- All user index.html files should be symlinked to the 'canonical' path so you don't need to re-run generate scripts to see changes

Cc: @jagtalon


bin/markup.pl
```diff
@@ -0,0 +1,33 @@
...
+    open my $fh, '<:encoding(UTF-8)', catfile( $build_dir, 'users.json' );
+    return <$fh>;
+}
+
+my $users = JSON::MaybeXS->new( utf8 => 1 )->decode( json );
+for my $user ( @{ $users } ) {
+    my $user_dir = catdir( $build_dir, $user );
+    make_path( $user_dir );
+    my $target = catfile( $user_dir, 'index.html' );
+    unlink $target;
```
MariagraziaAlastra: Here I get "Can't unlink('/home/ddgc/ddgc/ddh-userpages/championswimmer/index.html'): No such file or directory at bin/markup.pl line 29"

jbarrett: Thanks @MariagraziaAlastra - I've wrapped this in a check to ensure the file exists before nuking it.

We have to do it this way with the delete first since:

- There are existing index.html files generated from templates which need to be overwritten.
- There's no option in perl `symlink` (that I know of) to "force" link generation.

jagtalon:
:+1:

MariagraziaAlastra:
@jbarrett When running bin/generate.pl: "Use of uninitialized value $if_modified_since in substitution (s///) at /home/ddgc/community-platform/bin/../lib/DDGC/Web.pm line 334."

MariagraziaAlastra:
Oh well, it was probably a minor issue anyway. Can be solved in a different PR

jagtalon:
My bad I thought this was assigned to me. Sorry, @MariagraziaAlastra

MariagraziaAlastra:
Np! :)

PR #1298, opened 2016-03-31
Only filter if_modified_since once we've established it is populated

jbarrett:
Issue reported in #1297


MariagraziaAlastra:
Thanks @jbarrett! :+1:

PR #1301, opened 2016-03-31
IA Pages and Dev Pipeline - update maintainer and contributors URLs

MariagraziaAlastra:
Maintainer URL updated both in the pipeline sidebar and in the single IA Page.
TODO: contributors.

### DIFF
--- a/src/js/ia/DDH.js
+++ b/src/js/ia/DDH.js
@@ -7,5 +7,19 @@
 
     };
 
+    env.Utils = {
+        // Called on userpages links click
+        sendReq: function($obj) {
+            var username = $.trim($obj.text());
+            var id = $obj.attr("id").match(/maintainer/)? "maintainer" : "contributor";
+            var url_suffix = "?" + Math.ceil(Math.random() * 1e7);
+            var img = new Image();
+
+            username = username? "_" + username.replace(/[^0-9a-zA-Z]/g, "") : "";
+
+            img.src = "https://duckduckgo.com/t/iaptu_" + id + username + url_suffix;
+        }
+
+    };
 
 })(window);
--- a/src/js/ia/DDH.js
+++ b/src/js/ia/DDH.js
@@ -11,10 +11,11 @@
         // Called on userpages links click
         sendReq: function($obj) {
             var username = $.trim($obj.text());
-            var id = $obj.attr("id").match(/maintainer/)? "maintainer" : "contributor";
+            var id = $obj.attr("id");
             var url_suffix = "?" + Math.ceil(Math.random() * 1e7);
             var img = new Image();
 
+            id = (id && id.match(/maintainer/))? "maintainer" : "contributor";
             username = username? "_" + username.replace(/[^0-9a-zA-Z]/g, "") : "";
 
             img.src = "https://duckduckgo.com/t/iaptu_" + id + username + url_suffix;
--- a/src/js/ia/IAPage.js
+++ b/src/js/ia/IAPage.js
@@ -362,7 +362,7 @@
                     });
 
                     $("body").on('click', ".userpage-link", function(evt) {
-                        sendReq($(this).parent());
+                        Utils.sendReq($(this).parent());
                     });
 
                     $('body').on('click', '.switch.js-switch', function(evt) {
@@ -1525,18 +1525,6 @@
                                 
                             ia_data.staged.beta = 1;
                         });
-                    }
-
-                    // Called on userpages links click
-                    function sendReq($obj) {
-                        var username = $.trim($obj.text());
-                        var id = $obj.attr("id").match(/maintainer/)? "maintainer" : "contributor";
-                        var url_suffix = "?" + Math.ceil(Math.random() * 1e7);
-                        var img = new Image();
-
-                        username = username? "_" + username.replace(/[^0-9a-zA-Z]/g, "") : "";
-
-                        img.src = "https://duckduckgo.com/t/iaptu_" + id + username + url_suffix;
                     }
 
                     // Saves values for editable fields on the dev page
--- a/src/js/ia/IAPage.js
+++ b/src/js/ia/IAPage.js
@@ -32,6 +32,8 @@
                 //console.log("for ia id '%s'", DDH_iaid);
 
                 $.getJSON(json_url, function(ia_data) {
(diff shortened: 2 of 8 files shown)


MariagraziaAlastra:
This is done, but it can't be merged yet! :smile:

src/js/ia/IAPage.js
```diff
@@ -360,6 +361,10 @@
                         }
                     });
 
+                    $("body").on('click', ".userpage-link a", function(evt) {
```
bsstoner: Why `$("body").on(...)` for click handlers? Can't you bind directly to the links?
```
$(".userpage-link a").on('click', function(){})
```

src/templates/devinfo.handlebars
```diff
@@ -160,10 +160,10 @@
         {{/ne_and}}
     {{/if}}
     </h3>
-    <ul>
+    <ul id="contributors--readonly" class="userpage-link">
```
bsstoner: Why is `userpage-link` on the parent element of the link? Why not put it directly on the `<a>`?

src/js/ia/IAPage.js
```diff
@@ -1522,6 +1527,19 @@
...
+                    // Called on userpages links click
+                    function sendReq($obj) {
+                        var username = $.trim($obj.text());
+                        var id = $obj.attr("id").match(/maintainer/)? "maintainer" : "contributor";
+                        var url_suffix = "?" + Math.ceil(Math.random() * 1e7);
+                        var img = new Image();
+
+                        username = username? "_" + username : "";
+
+                        $obj.append(img);
```
bsstoner: Do you need to append it to the DOM? I thought setting `img.src` worked in the other PR?

src/js/ia/IAPage.js
```diff
@@ -1522,6 +1527,19 @@
...
                     }
 
+                    // Called on userpages links click
+                    function sendReq($obj) {
+                        var username = $.trim($obj.text());
+                        var id = $obj.attr("id").match(/maintainer/)? "maintainer" : "contributor";
+                        var url_suffix = "?" + Math.ceil(Math.random() * 1e7);
+                        var img = new Image();
+
+                        username = username? "_" + username : "";
```
bsstoner: Do you want to strip special characters or anything from the username? Or are they safe?

bsstoner:
I left some feedback, let me know if you want to discuss tomorrow.

src/js/ia/IAPage.js
MariagraziaAlastra: @bsstoner Most (if not all) of the IA Pages elements are dynamically created: after page load, we call [this function](https://github.com/duckduckgo/community-platform/blob/maria/replace-urls/src/js/ia/IAPage.js#L1810) which renders the Handlebars templates associated with each piece of the page in the desired order. This function is also used to refresh the page when editing a field.

This means binding the events to the elements directly [won't work](http://stackoverflow.com/questions/13650261/jquery-cannot-bind-events-to-dynamic-elements), since they previously didn't exist, while binding on body and then seeing if the specific element clicked is the right one works.

Something I learned in my early web dev days :smile:

src/js/ia/IAPage.js
MariagraziaAlastra: @bsstoner Oh, I added that because it didn't seem to work without it, previously, but I just tried commenting that out and looks like it wasn't necessary! 
I was probably doing something else wrong, then :smile: 

Just updated the code!

Thanks!

src/js/ia/IAPage.js
MariagraziaAlastra: @bsstoner Right, I forgot about that :octocat: 
Just updated the code, thanks!

src/templates/devinfo.handlebars
MariagraziaAlastra: @bsstoner No specific reason other than I did a bit of back and forth with that end ended up leaving the class in the parent element. Updated, thanks!

src/js/ia/IAPage.js
bsstoner: Okay, this is something way bigger than we should get into here, I figured there was probably a reason. :)

A more modular lifecycle pattern that frameworks typically use is something like 1) create view 2) render elements 3) bind events to elements 4) remove events 5) remove elements 6) destroy view.

src/js/ia/DDH.js
```diff
@@ -7,5 +7,20 @@
 
     };
 
+    env.Utils = {
+        // Called on userpages links click
+        sendReq: function($obj) {
```
bsstoner: It might be clearer to others in the future if this were called `sendPixel`? Up to you, not worth holding up if you want to merge.

bsstoner:
:+1: Looks good to me! Feel free to merge if you're ready.

MariagraziaAlastra:
@bsstoner Thanks! @jbarrett do you want to have a look here? Check links and general well-functioning?

jbarrett:
:+1: Ready to merge?

src/templates/commit_page.handlebars
```diff
@@ -193,14 +193,14 @@
                 <td name="maintainer"  id="maintainer_original">
                     {{#if original.maintainer}}
                         {{#if original.maintainer.github}}
-                            <a href="https://github.com/{{original.maintainer.github}}">{{original.maintainer.github}}</a>
+                            <a href="https://duckduckhack.com/u/{{slug original.maintainer.github}}">{{original.maintainer.github}}</a>
```
GuiltyDolphin: @MariagraziaAlastra We don't have `https` *yet*, but I think we will after #1330!

MariagraziaAlastra:
Moar changes! Now you can only see the links to userpages when you're logged in, as requested.

Test by opening single IA pages while logged out and then while logged in.
Expected: when logged out the links will redirect to github; when logged in, they'll redirect to userpages.

Same on the dev pipeline: select an IA in order to open the detail sidebar. When you're logged out the maintainer and contributors links will go to github, while when logged in they'll go to userpages.

src/templates/commit_page.handlebars
MariagraziaAlastra: @GuiltyDolphin yo thanks for spotting this :)

jdorweiler:
:+1:

PR #1305, opened 2016-04-01
Userpages - turn issues and pr arrays into hashes (prevents duplication)

MariagraziaAlastra:
//cc @jagtalon 
screenshot-maria duckduckgo com 5002 2016-04-01 14-55-05


jagtalon:
:+1:

PR #1306, opened 2016-04-01
Create symlinks for mixed-case logins

jbarrett:


jbarrett:
@MariagraziaAlastra With this change you should see a page for e.g. 'mariagraziaalastra' and 'MariagraziaAlastra'.

Thanks!

MariagraziaAlastra:
Yep, works as expected, thanks! :+1:

PR #1308, opened 2016-04-01
Gather.pm - can't use usernames as keys

MariagraziaAlastra:
//cc @jagtalon 
screenshot-maria duckduckgo com 5002 2016-04-01 21-49-49


jagtalon:
<3

PR #1313, opened 2016-04-05
Sync github users where we are missing them

jbarrett:


jbarrett:
Data gathering now requires an GitHub API key in your environment. If you haven't configured this you can run it like so:

```
$ DDGC_GITHUB_TOKEN=your_api_key_here bin/generate.pl
```

jagtalon:
:+1:

PR #1314, opened 2016-04-05
ghIssues updates

jdorweiler:
1. Only run ghIssues with the dax token in production.  You need to provide a token otherwise
2. There is a case where the old dax comment wan't correctly found (line: 487)


jdorweiler:
I also updated it to only get the last 6 hours of issues.  It runs every 30 minuets so this is plenty of overlap.

MariagraziaAlastra:
Thanks @jdorweiler, looks good :+1:

PR #1318, opened 2016-04-08
update file links

jdorweiler:
We moved all the longtail and fathead files from share to lib so the IA pages links are broken now.
https://github.com/duckduckgo/zeroclickinfo-longtail/pull/51#issuecomment-207578087

We also have some other broken links due to a bug in ghIssues that I recently fixed.
https://github.com/duckduckgo/community-platform/pull/1315

This will update all those broken links.

@MariagraziaAlastra @zachthompson


zachthompson:
Good, thanks.

PR #1320, opened 2016-04-11
IAIndex.js: Remove //g flag from regexp.

jagtalon:
We're using a regexp with a //g flag and running it on multiple strings. Since that regexp maintains state on subsequent calls (lastIndex), we run into matching problems. The solution is to just not use the g flag.


MariagraziaAlastra:
@jagtalon Thanks, LGTM :+1:

PR #1322, opened 2016-04-14
Dev Pipeline - fix dates in comments

MariagraziaAlastra:
On production the dates in the detail sidebar all show a negative number.
Turns out the timeago helper got passed an unneeded argument, since these dates come from GitHub and not from the beta server.
screenshot-maria duckduckgo com 5001 2016-04-14 14-48-09


russellholt:
can you add an explanation for the effect of that argument?

MariagraziaAlastra:
That argument is for when the dates come from the beta server, and since they're formatted differently, they need to be parsed differently - I added a comment in the code.
Thanks!

jbarrett:
:+1: Issue appears to be resolved with this change

PR #1324, opened 2016-04-15
UserPage.js - use login instead of id

MariagraziaAlastra:
This way it's more readable.


jbarrett:
:+1: Thanks!

MariagraziaAlastra:
:cactus:

PR #1329, opened 2016-04-19
Basic IA controller tests

jbarrett:
Cc: @MariagraziaAlastra @russellholt 

Here is a very basic IA flow (create a new IA, verify it exists in database) which should demonstrate:

- DB schema rollout
- Plack test app creation (with testutils session integration)
- How to create and get a session for a user of a given class.
- How to test things with CSRF mitigation.
- The most basic of [Test::More](https://metacpan.org/pod/Test::More) assertions (read this doc).

Thanks to @MariagraziaAlastra for help navigating the instant_answer controller.

### DIFF
--- a/t/instant_answer_basics.t
+++ b/t/instant_answer_basics.t
@@ -119,7 +119,7 @@ test_psgi $app => sub {
 
     # Check all_milestones JSON (should have everything)
     my $ia_repo = $get_repo_json->({ repo => 'longtail', all_milestones => 1 });
-    is( ref $ia_repo->{test_ia}, 'HASH', 'Test IA approved - is in all milestone repo JSON' );
+    is( ref $ia_repo->{test_ia}, 'HASH', 'Test IA is in all milestone repo JSON' );
 
     # "Approve" IA
     my $ia_approve_result = $set_ia_value->({ cookie => $cookie, public => 1 });
--- a/t/instant_answer_basics.t
+++ b/t/instant_answer_basics.t
@@ -118,18 +118,9 @@ test_psgi $app => sub {
     ok( $ia, 'Query returns something' );
     is( ref $ia, 'DDGC::DB::Result::InstantAnswer' ,'IA is a Result::InstantAnswer' );
 
-    # "Approve" IA
-    my $ia_approve_request = $cb->(
-        POST '/ia/save',
-        Cookie  => $cookie,
-        Content => [
-            id           => 'test_ia',
-            field        => 'public',
-            value        => 1,
-            action_token => $action_token,
-        ],
-    );
-    ok( $ia_approve_request->is_success, 'IA "Approve" request succeeds' );
+    # Check all_milestones JSON (should have everything)
+    my $ia_repo = $get_repo_json->({ repo => 'longtail', all_milestones => 1 });
+    is( ref $ia_repo->{test_ia}, 'HASH', 'Test IA approved - is in all milestone repo JSON' );
 
     is( $ia->public, 0, 'IA cannot be approved by non-admin' );
 
--- a/t/instant_answer_basics.t
+++ b/t/instant_answer_basics.t
@@ -129,6 +129,18 @@ test_psgi $app => sub {
     $ia->discard_changes;
     is( $ia->public, 1, 'IA Result is now public' );
 
+    # Check repo JSON - test_ia should be absent due to its milestone
+    $ia_repo = $get_repo_json->({ repo => 'longtail' });
+    ok( !$ia_repo->{test_ia}, 'Test IA not in published milestone' );
+
+    # Set IA milestone
+    my $ia_milestone_result = $set_ia_value->({ cookie => $cookie, dev_milestone => 'testing' });
+    is( $ia_milestone_result->{saved}, 1, 'Set test_ia milestone to testing' );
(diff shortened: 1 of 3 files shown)


jbarrett:
`prove -Ilib t/instant_answer_basics.t`

t/instant_answer_basics.t
```diff
@@ -0,0 +1,108 @@
...
+            action_token => $action_token,
+        ],
+    );
+    ok( $ia_approve_request->is_success, 'IA "Approve" request succeeds' );
+
+    # IA should be in repo JSON
+    my $ia_repo_json_request = $cb->( GET '/ia/repo/all/json?all_milestones=1' );
+    ok( $ia_repo_json_request->is_success, 'Can retrieve longtail JSON' );
+    my $ia_repo = decode_json( $ia_repo_json_request->decoded_content );
+    is( ref $ia_repo->{test_ia}, 'HASH', 'Test IA approved - is in repo JSON' );
```
MariagraziaAlastra: I wonder if this part should be taken out and generalized in an external sub.
Eg, I can imagine using it to verify whether changing dev milestone to an IA will make it show up / disappear from the repo JSON, and the "all_milestones" parameter could be passed in as a boolean.

I don't know if I'm explaining myself clearly :smile:

nilnilnil:
nice! @MariagraziaAlastra can you take this one?

t/instant_answer_basics.t
```diff
@@ -0,0 +1,110 @@
...
+        { username => 'ia_user' }
+    );
+    ok( $user_request->is_success, 'Creating an IA user' );
+
+    # Get session
+    my $session_request = $cb->(
+        POST '/testutils/user_session',
+        { username => 'ia_user' }
+    );
+    ok( $session_request->is_success, 'Getting IA user Cookie' );
```
nilnilnil: This `is_success` function seems pretty convenient :)

nilnilnil:
This is a great start @jbarrett, thanks!

MariagraziaAlastra:
@nilnilnil Sure, I'll take it :smile:

t/instant_answer_basics.t
jbarrett: Hmm, the specific test on L101 should actually be fixed and negated - since the current behaviour of `InstantAnswer::save_edit` is (I think) to return HTTP 200 + status data in JSON, we should be examining that.

Should the user who created the IA be able to set its public flag?

If we need to do something more than once, it should absolutely be extracted to a sub. I'm going to make some changes here anyway, since I am asserting the wrong thing.

t/instant_answer_basics.t
jbarrett: We get a [HTTP::Response](https://metacpan.org/pod/HTTP::Response) here - chock full of useful stuff.

nilnilnil: So that requires the backend sends something other than a 200 on a failed action?

jbarrett: Yep, explicitly 200 hence the issue here : https://github.com/duckduckgo/community-platform/pull/1329#discussion_r60245513

nilnilnil: Ah, derp. Thanks!

t/instant_answer_basics.t
```diff
@@ -0,0 +1,137 @@
...
+        my $u = URI->new("/ia/repo/$repo/json");
+        $u->query_form( all_milestones => $opts->{all_milestones} || 0 );
+
+        my $ia_repo_json_request = $cb->( GET $u->canonical );
+        ok( $ia_repo_json_request->is_success, "Can retrieve $repo JSON" );
+        return decode_json( $ia_repo_json_request->decoded_content );
+    };
+
+    my $get_action_token = sub {
+        my $action_token = $cb->(
```
jbarrett: @MariagraziaAlastra The reason I went with coderefs like this is stuff in the `test_psgi` scope (like `$cb`) is useful, so we don't have to pass around arbitrary stuff from this scope to external subs.

Thoughts?

MariagraziaAlastra: @jbarrett Sounds good, thanks for explaining :smile:

t/instant_answer_basics.t
```diff
@@ -0,0 +1,165 @@
...
+my $ia = $d->rs('InstantAnswer')->find('test_ia');
+
+$ia->update({ dev_milestone => 'live', perl_module => 'DDG::Longtail::TestIA' });
+$ia->add_to_topics( { name => 'interesting_topic' } );
+
+$mech->get_ok('/ia');
+$mech->content_contains('test_ia');
+
+$mech->get_ok('/ia/view/test_ia');
+$mech->title_is('Test Ia'); # Title case filter
```
jbarrett: This is an alternative approach to [this page content check in blog.t](https://github.com/duckduckgo/community-platform/blob/cc7a22b/t/blog.t#L90-L93).

[Test::WWW::Mechanize::PSGI](https://metacpan.org/pod/Test::WWW::Mechanize::PSGI) can completely replace or augment Plack::Test depending on your preference or which types of tests you're going to perform.

MariagraziaAlastra:
@jbarrett Thanks, this LGTM! :+1:

PR #1335, opened 2016-04-20
Userpages - check if ddgc accounts have a github account linked

### LINKED ISSUE #1331: Developed IA's are wrong on Contributor Profile
Lo,

On the new super-cool contributor profile it appears that I'm missing some IA's:
For me the following are appearing in Maintained but not developed:
* Bash Primary Expressions; 
* Conversions; 
* Regex Cheatsheet; 
* IPLookup; 
* Timediff (which is still in progress); 
* Parcelforce
* BloodType
* UrlDecode

@alohaas says that the Developed tab should be anything I've contributed to which would probably also include:
* ColorCodes
* ResistorColors
* PeopleInSpace

However there are others that I've done more minor changes to (Tips for example), that also don't currently appear. I'm sure there are others, but their names escape me currently! 

I would presume that anything that I'm currently a maintainer for on https://duck.co/ia?repo=goodies should appear in the maintainers list, which doesn't seem to be the case. For example I'm missing 
Moon 

cc @MariagraziaAlastra

MariagraziaAlastra:
Fixes #1331 
To test this, you can create a user with the same name as someone on duck.co who's listed as contributor on IA Pages, and then link to your github account.

Run generate.pm and see if in your userpage you get the IAs that user has contributed to.

### DIFF
--- a/lib/DDH/UserPage/Gather.pm
+++ b/lib/DDH/UserPage/Gather.pm
@@ -143,6 +143,18 @@ sub get_avatar {
     }
 }
 
+sub add_to_contributors {
+    my ( $self, $login, $ia, $ia_id, @contributors ) = @_;
+
+    my %contributor = (
+        username => $login,
+        avatar_url => $self->get_avatar($login)
+    );
+
+    push @{ $ia->{$ia_id}->{contributors} }, \%contributor;
+    push @contributors, $login;
+}
+
 sub transform {
     my ( $self, $ia ) = @_;
     my $transform = {};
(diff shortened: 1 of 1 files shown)


lib/DDH/UserPage/Gather.pm
```diff
@@ -167,13 +179,17 @@ sub transform {
...
-                            avatar_url => $self->get_avatar($login)
-                        );
+                        $self->add_to_contributors($login, $ia, $ia_id, @contributors);
+                    } elsif ( $developer->{type} && 
+                              $developer->{type} eq 'duck.co' ) {
+
+                            my $complat_user = $self->ddgc->rs('User')->find({ username => $developer->{name} });
+
+                            if ( $complat_user && ( my $gh_id = $complat_user->github_id ) ) {
+                                my $gh_user = $self->ddgc->rs('GitHub::User')->find({ github_id => $gh_id });
```
jbarrett: Do we want to check the `github_user_plaintext` column for people who haven't yet logged in with GH?

Down side of this is that data is completely unverified, but we generally act on good faith with people who have active contributions, so may not be a biggie?

MariagraziaAlastra: Yeah we should do that! Alright, thanks!

russellholt: Don't we want to update the database rather than ask the question here? Instead of code complexity to deal with data, fix the data.

jbarrett:
Data fixes incoming - thanks @MariagraziaAlastra!

PR #1338, opened 2016-04-20
fix traffic graph

jdorweiler:
The previous dbix query wasn't exactly getting us the right data for the pixel graphs.  It also needed an awkward `get_array_by_pixel` helper function.  So why not just do it all in the sql query, right?

I also update the date calculation to get us 30 days instead of the 28 days we're getting right now. 

Some before and after examples:
https://duck.co/ia/view/latex
Before
selection_334
After
selection_333

https://duck.co/ia/view/airlines
Before
selection_332

After
selection_331


russellholt:
Nice. Good to finally know the reason behind those strange zeros.

MariagraziaAlastra:
@jdorweiler Looks good, thanks! :+1:

PR #1339, opened 2016-04-20
Add frontend integration tests 

MariagraziaAlastra:
//cc @nilnilnil 
This is still very very WIP, I just started with a little demo of what we can do with casperjs.
TODO: add some modules with utility functions we can reuse, add grunt task.

**How to install casperjs on your instance:**
- `npm install` This will install phantomjs - I had to pin it to 1.9.8 because that's the latest release working seamlessly with casperjs

- `bower install` this will install casperjs, since it's available as a bower module.

- `export PATH=$PATH:node_modules/phantomjs/lib/phantom/bin`

**How to run a test:**
`casperjs test [filename] --hostname=[hostname]`

For example, in my case:  `casperjs test src/t/ia/index.js --hostname=maria`.

When we'll have the grunt task it should be even easier, though.

Thanks @jbarrett for the help!

new-blank

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ install:
   - dzil authordeps | grep -ve '^\W' | xargs -n 5 -P 10 cpanm  --quiet   --notest  --skip-installed --mirror http://www.cpan.org/ --mirror http://duckpan.org/
   - dzil listdeps | grep -ve '^\W' | cpanm  --quiet   --notest  --skip-installed --mirror http://www.cpan.org/ --mirror http://duckpan.org/
   - npm install grunt --save-dev
+  - npm install load-grunt-tasks
   - npm install -g grunt-cli casperjs
 language: perl
 perl:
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -28,3 +28,4 @@ perl:
   - 5.16
 script:
   - dzil smoke --release --author
+  - grunt test-ia
```
jbarrett: We need to figure out how to install the dependencies for this. It will probably require some experimentation.

We can probably use node and npm from apt? Then add grunt and bower commands to the install block?

Also need to make sure phantomjs is in the `$PATH`

jbarrett:
@MariagraziaAlastra Would appreciate it if you could review my changes here.

Happy to merge when you are.

Thanks!

MariagraziaAlastra:
@jbarrett Looks really good, thanks a lot for solving the Travis issue! :smile: :+1:

PR #1340, opened 2016-04-21
Small script wrapping Devel::Cover

jbarrett:
This is a tiny but hopefully useful script you can use to get test coverage metrics. The final result is a set of html in cover_db/

You can use something like [App::HTTPThis](https://metacpan.org/pod/App::HTTPThis) to quickly serve your current directory.

You'll also need to install [Devel::Cover](https://metacpan.org/pod/Devel::Cover)


jbarrett:
WIP until Pinto updated.

PR #1346, opened 2016-04-27
Licensing

aanarres:
This PR is a first pass to document licenses for community-platform project.

Changes:
- added license to root directory. I added GPLv1 or later or Artistic 1.0, because I found it in dist.ini and it seems the intended license for the project, where no other is specified.
- added license to /src. I found in bower.json that the intended license for js and visual stuff in /src is MIT.
- added sections to README about licensing.
- reviewed all project and found third party components, added information about their licenses.

Please advise if the intentions were different than what I found.

There are some things left to do, like:
- images, or are all images in community-platform supposed to be covered by Artistic 1.0/GPL 1 or later licenses?
- fonts (/root/static/fonts/, /share/annifont.ttf), are there licenses for fonts? are there 3rd party fonts or were they designed in-house? I found SIL OFT at a cursory search, but I didn't review fonts yet.
- in lib/DDGCTest/Database.pm, there's a note about wallpapers being CC-BY-NC-SA 3.0, I don't know what wallpapers is that text talking about, are they in a repository on github?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
-This software is copyright (c) 2011 by Gabriel Weinberg <yegg@duckduckgo.com>,
-DuckDuckGo Community <community@duckduckgo.com>
+This software is copyright (c) 2011 by DuckDuckGo, Inc. http://duckduckgo.com,
+DuckDuckGo Community
 
 This is free software; you can redistribute it and/or modify it under
 the same terms as the Perl 5 programming language system itself.
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-This software is copyright (c) 2011 by DuckDuckGo, Inc. http://duckduckgo.com,
+This software is copyright (c) 2011 by DuckDuckGo, Inc. https://duckduckgo.com,
 DuckDuckGo Community
 
 This is free software; you can redistribute it and/or modify it under
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,379 @@
+This software is copyright (c) 2011 by Gabriel Weinberg <yegg@duckduckgo.com>,
+DuckDuckGo Community <community@duckduckgo.com>
+
+This is free software; you can redistribute it and/or modify it under
+the same terms as the Perl 5 programming language system itself.
+
+Terms of the Perl programming language system itself
+
+a) the GNU General Public License as published by the Free
+   Software Foundation; either version 1, or (at your option) any
+   later version, or
+b) the "Artistic License"
+
+--- The GNU General Public License, Version 1, February 1989 ---
+
+This software is Copyright (c) 2011 by Gabriel Weinberg, DuckDuckGo Community
+
+This is free software, licensed under:
+ 
+  The GNU General Public License, Version 1, February 1989
+ 
+                    GNU GENERAL PUBLIC LICENSE
+                     Version 1, February 1989
+ 
+ Copyright (C) 1989 Free Software Foundation, Inc.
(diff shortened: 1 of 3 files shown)


jbarrett:
Thanks @aanarres - I will get back to you on this as soon as I can (Tuesday at the latest)

Very sorry for the delay in responding!

LICENSE
```diff
@@ -0,0 +1,379 @@
+This software is copyright (c) 2011 by DuckDuckGo, Inc. http://duckduckgo.com,
```
GuiltyDolphin: Zeh `https`.

aanarres: Fixed, thanks! I must have copied it from [dist.ini](https://github.com/duckduckgo/community-platform/blob/master/dist.ini). Do you want it fixed there too?

tagawa: Hi @aanarres. Thanks for checking. I've just submitted a PR (https://github.com/duckduckgo/community-platform/pull/1374) for `dist.ini` so that your PR can focus on licensing.

tagawa:
This all looks good to me but given that it's legal text, I'd like to get further review. /cc @jbarrett

alohaas:
ping @jbarrett ! any input?

src/LICENSE
```diff
@@ -0,0 +1,19 @@
+The MIT License (MIT)
+Copyright (c) <year> <copyright holders>
```
GuiltyDolphin: Looks like the year and holder hasn't been updated here.

alohaas: @GuiltyDolphin good catch.

aanarres: @GuiltyDolphin you're right, it's better to fill these blanks. I committed a change where (c) holders are credited similarly with another part of community-platform. Please do tell if it's reasonable enough or something else should be mentioned.

GuiltyDolphin: @aanarres Oh, I can't really comment - I'm not a member of the platform; I'm just a bit nosey :wink:

nilnilnil:
I have a question. We've preferred using Apache 2.0 because of the trademarks clause. We want to develop free software! We don't however want that software redistributed under our brand.

I don't see any particular protection regarding that here. What can we do about that?

aanarres:
@nilnilnil Well, there isn't a need for a copyright license to talk about trademarks. Many projects manage these concerns separately. Trademarks need protected on their own anyway, no matter if the copyright license says there are trademarks or not.

There are cases of projects licensed under an open source license that doesn't mention trademarks, and where the project's stewards had trademarks on names in the source, who have required someone who forked the project to stop using their trademarks when distributing. Of course, successfully. I don't know of any case who hasn't been successful for this reason.

That said, there are licenses that do indeed include "just in case" text in the copyright license. Some go further than Apache 2.0 even, and add clauses like "Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission" ([BSD](https://opensource.org/licenses/BSD-3-Clause)), which is actually a trademark-like restriction.

To address that you feel important to except them explicitly, I suggest to add text to this effect in another file of the project. Can be README file, or can be a new file that will contain the credits and licensing of included components and a statement on trademarks.

There are other ways:
- add a clause to existing licenses. I can't recommend this option, because this concern can be addressed otherwise than creating a new license. A new license is something that everyone else will then have to interpret instead of relying directly on the known names of existing licenses;
- relicense with a license mentioning trademarks. I don't know if it's wanted or if it's easy to do, it depends what contributions are in this project.

nilnilnil:
@aanarres thanks for the follow-up. As you can tell, we talked about it a bit, and think this is the right move. Thanks for your thoroughness.

PR #1352, opened 2016-05-10
Gather.pm - list other issues and prs as well

MariagraziaAlastra:
//cc @jagtalon 

Need to test this more


russellholt:
👍 how WIP is this? what's the next step?

MariagraziaAlastra:
Just testing

jbarrett:
I think we should develop some unit tests here. We can tack new data creation onto [instant_answer_basics.t](https://github.com/duckduckgo/community-platform/blob/698f09d/t/instant_answer_basics.t), then Gather, and pass a `tempdir()` or something as the target for Generate.

I'm not sure how to make this build known to other processes, but once we figure that out we can make another version of [run.bash](https://github.com/duckduckgo/community-platform/blob/698f09d/src/t/run.bash) to invoke [bin/dev_server.psgi](https://github.com/duckduckgo/community-platform/blob/698f09d/bin/dev_server.psgi).

jbarrett:
I can jump into this if there are no objections.

MariagraziaAlastra:
@jbarrett Sure, thanks!

jbarrett:
Sorry for --force push here - merged master into dev-ddh and rebased this branch off that.

jbarrett:
So these changes are just skeletal - they don't do anything since the appropriate data and tests aren't in place.

I think we need to mock data since relying on Github's state can't be relied upon.

Anyway, next steps are to create IA and user data which will result in JSON files being generated. There's a script to launch the userpage static file server too, but it's up to you if we want to add casper tests right now - might need a little more work to get this going, kinda beyond the scope of this one task.

jbarrett:
We need to something about these phantomjs build failures.

It looks like there's an instance installed as part of the node setup:

https://travis-ci.org/duckduckgo/community-platform/builds/129759876#L56

Can we make use of this?

MariagraziaAlastra:
@jbarrett Should I remove phantomjs from package.json then?

jbarrett:
@MariagraziaAlastra Not just yet, thanks - will figure it out outside this task.

We need it in package.json for local testing anyway, right?

MariagraziaAlastra:
@jbarrett Yeah, correct

jbarrett:
@MariagraziaAlastra Please review these changes. Thanks!

MariagraziaAlastra:
@jbarrett LGTM, thanks! :+1:

PR #1353, opened 2016-05-11
Gather.pm - count_prs and count_issues should only include open ones

MariagraziaAlastra:
//cc @jagtalon 

Currently, they include closed/merged issues and prs as well, but we only want to know how many are open.


russellholt:
LGTM

jbarrett:
@russellholt We have no tests here - how does it LG?

russellholt:
TM

PR #1355, opened 2016-05-13
Add tests for #1353

MariagraziaAlastra:
I created some dummy data to test prs_count and issues_count. They should contain the number of prs and issues with 'open' state and exclude closed ones.


jbarrett:
:+1: Many thanks for following up on this, @MariagraziaAlastra!

PR #1384, opened 2016-06-01
Style the banner

jagtalon:
Style the banner according to spec.

screen shot 2016-06-01 at 3 40 04 pm


jagtalon:
@MariagraziaAlastra sorry added the notification in here too:

screen shot 2016-06-01 at 3 40 04 pm

MariagraziaAlastra:
Thanks! :+1:

PR #1387, opened 2016-06-02
InstantAnswer.pm - call format_id() inside create_ia_from_pr()

MariagraziaAlastra:
##### Description :
IDs (and meta_ids) should always be case insensitive, but the format_id() function wasn't called in (here)[https://github.com/duckduckgo/community-platform/blob/master/lib/DDGC/Web/Controller/InstantAnswer.pm#L1259], which allowed people to use uppercase in their IDs when creating an IA Page from a PR.
Just added a call to that function in there, it was an easy fix.

##### Reviewer notes :
When creating an IA Page from the new IA wizard (ia/new_ia), both when filling in the fields and when skipping the wizard by using a PR link, the resulting meta_id for should always be all lowercase.

[This](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/3134) was the PR used to create the IA Page with the uppercase ID, but it's fixed now.

#

##### Who should be informed of this change?
@moollaza @tagawa 

##### Checklist :
- [x] Back end tests (perl, scripts)
- Browser verification
    - [ ] IE
    - [ ] Chrome
    - [x] Firefox
    - [ ] Safari
    - [ ] Opera 
- Mobile verification
    - [ ] iOS
    - [ ] Android

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -1357,6 +1357,8 @@ sub create_ia_from_pr :Chained('base') :PathPart('create_from_pr') :Args() {
                     
                     # Capitalize each word in the name string
                     $name =~ s/([\w']+)/\u\L$1/g;
+
+                    $id = format_id($id);
                     
                     my $new_ia = $c->d->rs('InstantAnswer')->update_or_create({
                         id => $id,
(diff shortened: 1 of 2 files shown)


MariagraziaAlastra:
Tests incoming!

MariagraziaAlastra:
Ready for review. Thanks @jbarrett for the help with tests :smile:

jbarrett:
:+1: LGTM, Thanks @MariagraziaAlastra!

CI is currently broken (known issue with Perl module versions) but tests also LGTM and pass in dev.

PR #1390, opened 2016-06-03
Use ngBind to replace the <title>

jagtalon:
Since JS needs to execute and fetch the data, the markup in the `<title>` tag is recorded in the history. To get around it, I used the ngBind directive to replace the text inside the tag once the data is available instead of using the template markup.


MariagraziaAlastra:
LGTM! :+1:

PR #1395, opened 2016-06-07
Fix overflow in the banner

jagtalon:


MariagraziaAlastra:
@jagtalon Looks like the banner is wider when I browse the dev pipeline, is this expected?
screenshot-maria duckduckgo com 5001 2016-06-08 13-30-38

jagtalon:
@MariagraziaAlastra That is of design since thy banner should match thy content's size.

MariagraziaAlastra:
LGTM :+1:

PR #1415, opened 2016-08-31
remove topics and add programming mission topics

jdorweiler:
Remove all the old topics and update with new programming mission search spaces. We're going to use topics to tag IAs as being part of the programming mission. 
I'm also adding the topics to stack_overflow since the topics update template only allows 3 topics. We might have to fix that later to allow more.


jdorweiler:
We can skip this for now. I have a different way to handle it without updating the database.

PR #1417, opened 2016-09-07
update dax message

jdorweiler:
One of the dax comments that we're looking for in ghIssues was updated. We need to update that in ghIssues too.


MariagraziaAlastra:
👍

PR #1418, opened 2016-09-07
update ia page regex

jdorweiler:
Updated the regex that finds IA page links in the first post of issues and pull requests. I updated it to be a little more forgiving by allowing extra spaces before the url and pulling off trailing spaces after the url.

cc/ @GuiltyDolphin

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -84,8 +84,8 @@ sub getIssues{
 			# Instant Answer Page: Link   <- preferred standard link
 			# [Instant Answer Page](Link) <- GHFM link
 			$issue->{'body'} =~ qr{
-				\[(?:Instant\s?Answer|IA)\sPage\]\(https?://duck\.co/ia/view/(?<id>\w+)\)
-				| (?:Instant\s?Answer|IA)\sPage:\s https?://duck\.co/ia/view/(?<id>\w+)}ix;
+				\[(?:Instant\s?Answer|IA)\sPage\]\(https?://duck\.co/ia/view/(?<id>\w+[^\s])\)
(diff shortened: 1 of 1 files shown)


MariagraziaAlastra:
Thanks @jdorweiler 👍

script/ghIssues.pl
```diff
@@ -84,8 +84,8 @@ sub getIssues{
 			# Instant Answer Page: Link   <- preferred standard link
 			# [Instant Answer Page](Link) <- GHFM link
 			$issue->{'body'} =~ qr{
-				\[(?:Instant\s?Answer|IA)\sPage\]\(https?://duck\.co/ia/view/(?<id>\w+)\)
-				| (?:Instant\s?Answer|IA)\sPage:\s https?://duck\.co/ia/view/(?<id>\w+)}ix;
+				\[(?:Instant\s?Answer|IA)\sPage\]\(https?://duck\.co/ia/view/(?<id>\w+[^\s])\)
+				| (?:Instant\s?Answer|IA)\sPage:?\s+ https?://duck\.co/ia/view/(?<id>\w+[^\s])}ix;
```
GuiltyDolphin: Why do we need the `[^\s]` at the end? Surely it'll catch this anyway?

Btw I think `\S` can be used as an equivalence of `[^\s]`.

jdorweiler: Yeah you're right we could do with out the `[^\s]`. It would be nice to break this out into a sub so we could do some unit testing though.

GuiltyDolphin: @jdorweiler I could pull it out if you want.

Are you thinking something along the lines of `sub issue_ia_page : IssueText -> IA Page`?

PR #1420, opened 2016-09-08
IA Pages - automatically update dev_milestone

MariagraziaAlastra:
##### Description :
Based on other fields values (perl_module, production_state, blockgroup, deployment_state, and the current dev_milestone) we should be able to automatically update the dev_milestone field for any IA.

This PR also introduces a new dev_milestone value: "offline".

How the dev_milestone will be determined:
   - online, live, blockgroup, perl_module, deployment_state = **live**
   - online, live, no perl_module/blockgroup/deployment_state = **offline**
   - offline, planning/development, no blockgroup, no deployment_state = **no change (still planning or development)**
   - offline, deprecated = **no change**
   - offline, complete, no deployment_state = **complete**
   - offline, planning/development, blockgroup = **testing**
   - offline, live, blockgroup, deployment_state = **offline**
   - offline, complete, blockgroup, deployment_state = **no change**

To do:
- [ ] Make sure all the possible cases are taken into account.
- [ ] SQL to automatically update existing IAs.
- [ ] Tweak the ghIssues.pl script to use the new statuscheck function when automatically updating the dev_milestone
- [ ] Back end tests

#

##### Checklist :
- [ ] Back end tests (perl, scripts)
- Browser verification
    - [ ] IE
    - [ ] Chrome
    - [ ] Firefox
    - [ ] Safari
    - [ ] Opera 
- Mobile verification
    - [ ] iOS
    - [ ] Android

### DIFF
--- a/lib/DDGC/Web/Controller/InstantAnswer.pm
+++ b/lib/DDGC/Web/Controller/InstantAnswer.pm
@@ -1116,6 +1116,22 @@ sub save_edit :Chained('base') :PathPart('save') :Args(0) {
                 save($c, \@update, $ia, 1);
                 $saved = 1;
                 
+                if ($field eq "production_state" || $field eq "deployment_state" || $field eq "dev_milestone"
+                   || $field eq "perl_module" || $field eq "blockgroup") {
+                    $ia->{$field} = $value;
+
+                    my $milestone = $ia->dev_milestone;
+                    my $perl_module = $ia->perl_module ? 1 : 0;
+                    my $online = ($ia->production_state eq "online") ? 1 : 0;
+                    my $blockgroup = $ia->blockgroup ? 1 : 0;
+                    my $deployment_state = $ia->deployment_state ? 1 : 0;
+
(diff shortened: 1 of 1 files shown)


lib/DDGC/Web/Controller/InstantAnswer.pm
```diff
@@ -1649,6 +1665,39 @@ sub update_ia {
...
+                $dev_milestone = "development";
+            } else if ($blockgroup) {
+                if ($deployment_state) {
+                    $dev_milestone = "complete";
+                } else {
+                    $dev_milestone = "testing";
+                }
+            }
+        } 
+    }
```
russellholt: logic nightmare

zachthompson:
> online, live, no perl_module/blockgroup/deployment_state = offline

Probably should be something other than "offline" since it indicates an inconsistent state.

zachthompson:
In general I don't think we are determining dev_milestone here.  Yes, we've used that column in the past for status but

1. It can be edited
2. It probably can and should be modifiable independently of the status, e.g. we *could* set it to dev when making an enhancement even though still online.

PR #1425, opened 2016-10-13
Hotfix: Remove last duck.co email addresses from templates / email

jbarrett:
MX config for this domain will be removed.


MariagraziaAlastra:
Thanks @jbarrett, LGTM! 👍

PR #1432, opened 2016-11-11
update live fatheads

jdorweiler:
##### Description :

There are many live fatheads that don't have the correct production state set. We're going to actually start using production state for fatheads. This sets all live fatheads to the correct production state.


MariagraziaAlastra:
Thanks @jdorweiler, this LGTM! 👍

PR #1437, opened 2016-12-05
encode spaces

### LINKED ISSUE #1436: Dax Metadata GitHub Post Broken: Markdown link urls need to be encoded
The automated message from Dax that posts to GitHub has broken and now shows raw markdown links for the example queries.

It looks like we are not properly URL encoding the query string to have `+` for spaces, so GitHub is not converting the text into a link.

This may be a new GitHub issue, but we can fix on our end by URL Encoding the queries

Here is a link to that example:

https://github.com/duckduckgo/zeroclickinfo-spice/pull/3061#issuecomment-263293898

jdorweiler:
##### Description :
Encode spaces in markdown links to fix #1436

### DIFF
--- a/script/ghIssues.pl
+++ b/script/ghIssues.pl
@@ -78,7 +78,6 @@ sub getIssues{
             $progress->update($line) unless $d->is_live;
             $line++;
 
-
             # get the IA name from the link in the first comment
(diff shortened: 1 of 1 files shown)


MariagraziaAlastra:
👍

PR #1438, opened 2016-12-13
Update subscription behaviour, content

jbarrett:
##### Description :

Changing verification rules


jbarrett:
Something else happening here

PR #1446, opened 2017-01-19
Emails Cleanup

thm:
This includes a bunch of cleanup in the emails, and updated verify and unsub templates.

Several images have yet to be updated in core. I expect there are some minor margin issues to solve after that.

@jbarrett 
cc @AdamSC1-ddg


jbarrett:
👍

PR #1507, opened 2018-03-08
Export help content

jbarrett:
##### Description :

This script should help export help content to files.

##### Reviewer notes :

Run the script. Help and images should be in ~/ddgc/help/$epoch directory.

Tarball for that timestamp should be in ~/ddgc/help/

#

##### Who should be informed of this change?

@MariagraziaAlastra @moollaza @alohaas 

##### Does this change have significant privacy, security, performance or deployment implications?

No

##### Checklist :
- [ ] Back end tests (perl, scripts)
- Browser verification
    - [ ] IE
    - [ ] Chrome
    - [ ] Firefox
    - [ ] Safari
    - [ ] Opera 
- Mobile verification
    - [ ] iOS
    - [ ] Android


script/ddgc_export_help.pl
```diff
@@ -0,0 +1,147 @@
...
+        my $src = $node->attr('src');
+        if ( $src =~ m{^https?://}i ) {
+            my $m;
+            try {
+                $m = $self->ddgc->rs('Media')->create_via_url( $self->user, $src );
+            } catch {
+                warn "Downloading $src failed : $_";
+            };
+            next unless $m;
+            $src = '/media/' . $m->filename;
```
MariagraziaAlastra: Is there any chance `$m` could have a value but not have the filename property, eg if the download succeeds but it's garbled up or the content isn't what we expect? 
I guess it's a pretty slim chance and every other possible error is handled, so doesn't matter

MariagraziaAlastra:
Left a small comment, LGTM and thanks for doing this! 👍

script/ddgc_export_help.pl
jbarrett: Should be OK - the [Media ResultSet's create call](https://github.com/duckduckgo/community-platform/blob/master/lib/DDGC/DB/ResultSet/Media.pm#L58) could probably use some guards 

Either way, the script is more likely to just crash than not have a filename - it's generated by all the md5 trash above that line and does not depend on the download's properties.

moollaza:
@jbarrett a few things I noticed while reviewing the content on the POC GH Page:
- Exported HTML contains `<html>` and `<body>` tags which we probably don't want in there
- Many lines end in `&#xD;` which doesn't seem to be rendered, but doesn't need to be in the content
- Neither of these seem to be breaking Jekyll though, which is great!

jbarrett:
We also need to change image URLs to be relative. Will get this done today.

MariagraziaAlastra:
Let's merge this since it's ready 👍

PR #1509, opened 2018-03-13
Fix mixed content issue

akalankauk:
http://www.youtube.com/embed/{0}?wmode=opaque'+'
http://www.youtube.com/embed/{id}?wmode=opaque

##### Description :
Update mixed content issue


jbarrett:
Thanks @akalankauk! I don't think we use this file. We should probably remove it from includes instead.

akalankauk:
Okay! 👍

PR #1514, opened 2018-04-11
Remove ref to blue-lightest

bsstoner:
We removed `$blue-lightest` here: https://github.com/duckduckgo/duckduckgo-colors/pull/6/files

This PR just removes the reference and creates the color locally here.

/cc @bbraithwaite


MariagraziaAlastra:
👍

